PR #5, opened 2022-09-10
Some fixes

hfiguiere:
Some more fixes.

### DIFF
--- a/flatpak_builder_lint/checks/modules.py
+++ b/flatpak_builder_lint/checks/modules.py
@@ -72,5 +72,3 @@ class ModuleCheck(Check):
         if modules := manifest.get("modules"):
             for module in modules:
                 self.check_module(module)
-        else:
-            self.errors.append("toplevel-no-modules")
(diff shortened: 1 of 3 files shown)


flatpak_builder_lint/checks/modules.py
```diff
@@ -2,35 +2,35 @@
 
 
 class ModuleCheck(Check):
-    type = "manifest"
+    type = "module"
```
hfiguiere, line 5: This is a reversal of your changes in commit b253e61 because the checker is called with a module not with a full manifest.

We could change this, but I found this is a nicer way to go at it.

hfiguiere: I could see it as having a type "manifest:module"

barthalion:
Looks like I implemented most of this in the morning as I didn't realize there's a PR open.

hfiguiere:
rebased. In the end put back a check you removed: the lack of top level `modules`.

hfiguiere:
Do we really require python 3.10 (as per the poetry file)? I know 3.8 at least with `:=`.

barthalion:
Nothing specific requires it now, but I don't see the point in targetting older versions.

PR #9, opened 2022-09-13
Flathub json check

hfiguiere:
The flathub.json check didn't work.

Also added check for architectures.

### DIFF
--- a/flatpak_builder_lint/checks/flathub_json.py
+++ b/flatpak_builder_lint/checks/flathub_json.py
@@ -1,4 +1,4 @@
-from . import Check
+from . import ARCHES, Check
 
 
 class FlathubJsonCheck(Check):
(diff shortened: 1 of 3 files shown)


hfiguiere:
3rd time is a charm

flatpak_builder_lint/checks/flathub_json.py
```diff
@@ -43,3 +45,25 @@ def check(self, manifest: dict) -> None:
                 if modules := manifest.get("modules"):
                     if self._check_if_extra_data(modules):
                         self.errors.append("flathub-json-modified-publish-delay")
+
+        if only_arches := flathub_json.get("only-arches"):
+            if "arm" in only_arches:
+                self.errors.append("flathub-json-deprecated-arm-arch-included")
+            if "i386" in only_arches:
+                self.errors.append("flathub-json-deprecated-i386-arch-included")
```
barthalion, line 53: These are not errors in practice.

flatpak_builder_lint/checks/flathub_json.py
```diff
@@ -43,3 +45,25 @@ def check(self, manifest: dict) -> None:
...
+            if len(only_arches) == 0:
+                self.errors.append("flathub-json-only-arches-empty")
+            if len(supported_arches.intersection(only_arches)) == len(supported_arches):
+                self.errors.append("flathub-json-redundant-only-arches")
+
+        if exclude_arches := flathub_json.get("exclude-arches"):
+            if "arm" in exclude_arches:
+                self.errors.append("flathub-json-deprecated-arm-arch-excluded")
+            if "i386" in exclude_arches:
+                self.errors.append("flathub-json-deprecated-i386-arch-excluded")
```
barthalion, line 63: Same here,

flatpak_builder_lint/checks/flathub_json.py
```diff
@@ -1,5 +1,7 @@
 from . import Check
 
+supported_arches = {"x86_64", "aarch64"}
```
barthalion, line 3: Does it make sense to make it a global variable? It could be moved to the class itself.

flatpak_builder_lint/checks/flathub_json.py
hfiguiere: but the goal here is to minimize the presence of `flathub.json`.

flatpak_builder_lint/checks/flathub_json.py
hfiguiere: actually it should be to some extent a global one, as we could check the same for the `only-arches` or `exclude-arches` in modules or sources.

Further down, we could have check with heuristic for errors like "you are using x86_64 binaries on aarch64".

I can't stop counting how often I have to mention this in reviews.

flatpak_builder_lint/checks/flathub_json.py
barthalion: Errors from this linter will eventually prevent new PRs from being merged. Complaining about deprecated arch used in flathub.json that has no effect on anything related to the build will only contribute to maintainers' frustration.

PR #19, opened 2022-10-04
Add RetroArch for wayland and x11 support

RobLoach:
The org.libretro.RetroArch package will fallback to wayland on non-x11 platforms, so have the linting flag that as an exception.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1118,6 +1118,9 @@
     "org.kde.kdevelop": {
 	"finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
+    "org.libretro.RetroArch": {
+        "finish-args-contains-both-x11-and-wayland": "fallbacks to wayland on non-x11 system"
+    },
     "org.mozilla.Thunderbird": {
         "finish-args-contains-both-x11-and-wayland": "fallbacks to wayland on non-x11 system"
     },


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1118,6 +1118,9 @@
     "org.kde.kdevelop": {
 	"finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
+    "org.libretro.RetroArch": {
+        "finish-args-contains-both-x11-and-wayland": "fallbacks to wayland on non-x11 system"
```
RobLoach, line 1122: I agree it's pretty poor. I had copied the description from Thunderbird's below this entry.

> fallbacks to wayland on non-x11 system

RetroArch has support for both wayland and x11. How about....

``` suggestion
        "finish-args-contains-both-x11-and-wayland": "falls back to x11 when wayland is not available"
```

hfiguiere: then you don't need the exception and just have `fallback-x11` instead of `x11` as the linter rule is about.

About t-bird, also had the same comment but it was merged anyway.

RobLoach: Ah, didn't know of fallback-x11. Thanks so much :+1:

RobLoach:
Thanks so much for the help! Adopting `fallback-x11` did it.

flatpak_builder_lint/staticfiles/exceptions.json
Erick555: @hfiguiere t-bird (or every mozilla app) is unique because it uses wayland as secondary choice (not primary as most apps do) only when no x11 socket (xwayland) is available. fallback-x11 artificially creates such environment but the developers intent is to use x11 even on wayland environment unless there is no choice.

PR #20, opened 2022-10-05
Add com.riverbankcomputing.PyQt.BaseApp to exceptions

JakobDev:


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1408,5 +1408,10 @@
     },
     "org.srb2.SRB2Kart": {
         "flathub-json-modified-publish-delay": "extra-data"
+    },
+    "com.riverbankcomputing.PyQt.BaseApp": {
+        "flathub-json-modified-publish-delay": "This is a BaseApp",
```
Erick555, line 1413: Why publish delay would be important for a BaseApp?

hfiguiere: it shouldn't have one.

JakobDev: https://github.com/flathub/com.riverbankcomputing.PyQt.BaseApp/pull/9

Erick555: I doubt this will be considered as sufficient explanation given flathub admins stance on (ab)using publish delay time. 2h more or less doesn't change much in this case.

barthalion: Yeah, that's poor excuse. There's nothing special about this BaseApp to have it a shorter publish delay.

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1408,5 +1408,10 @@
...
+        "flathub-json-skip-appstream-check": "This is a BaseApp",
+        "finish-args-not-defined": "This is a BaseApp"
```
barthalion, line 1415: I guess we should handle these exceptions in the linter code for all BaseApps. I will look into it.

barthalion: https://github.com/flathub/flatpak-builder-lint/commit/827794d0afc96afacaf2cbe4c1b136b7fcc36379 and https://github.com/flathub/flatpak-builder-lint/commit/4bfe5d874e1afdb93c69959e64c9b7226345de92 fix it.

flatpak_builder_lint/staticfiles/exceptions.json
JakobDev: Why do we need the delay at all? The Updates are tested before merging. So why should we keep the new version of the BaseApp, which may have changes/bugfixes that are really needed (I had such a case with one of my Apps) unpublished on the Buildserver for a few Hours. I just failed to understand the reason for this. There were no complaints about the shorter delay at all.

flatpak_builder_lint/staticfiles/exceptions.json
bbhtt: this was done later in https://github.com/flathub-infra/flatpak-builder-lint/commit/d47ac5c881f42eb34416006eb4c8a1d43362edf2 and https://github.com/flathub-infra/flatpak-builder-lint/commit/3efa25af3229ee2b89a7e8b7becd213e79ebd05e

Issue #23, opened 2022-10-07
Failure when checking org.freedesktop.fwupd.json


hughsie:
I just get a failed build and:

	{
	    "errors": [
		"toplevel-cleanup-debug",
		"flathub-json-skip-appstream-check"
	    ],
	    "warnings": [
		"toplevel-unecessary-branch",
		"toplevel-command-is-path"
	    ]
	}

This isn't super helpful; I literally don't know what to do.

hfiguiere:
> I just get a failed build and:
> 
> ```
> {
>     "errors": [
> 	"toplevel-cleanup-debug",

there is `/lib/debug` in cleanup. it shouldn't be this prevent to have proper debug info package generate.

> 	"flathub-json-skip-appstream-check"

appstream check in flathub.json. If justified, an exception needs to be added.

>     ],
>     "warnings": [
> 	"toplevel-unecessary-branch",

Don't set `branch` in the manifest for `stable`.

> 	"toplevel-command-is-path"

`command` shouldn't be a path., but a command that will be found in the path.

hfiguiere:
the last two are warnings.

hughsie:
Right, but it says none of that in the failure in flathub....

hughsie:
> appstream check in flathub.json. If justified, an exception needs to be added.

fwupd is a command line app, and I don't think Flathub wants to show CLI apps. Is this not true anymore?

hfiguiere:
> > appstream check in flathub.json. If justified, an exception needs to be added.
> 
> fwupd is a command line app, and I don't think Flathub wants to show CLI apps. Is this not true anymore?

This is a valid exception, definitely.

hughsie:
Are not all console apps on flathub valid exceptions?

hfiguiere:
they are definitely hidden, but they also can have a valid appstream file.

hfiguiere:
See #24

hughsie:
ohh, fwupd has a valid metainfo file

hughsie:
If it has a metainfo.xml, do i still need the flathub.json?

hfiguiere:
If it validates, no you don't need the exception. That's the best option

hughsie:
> If it validates...

https://github.com/flathub/org.flatpak.Builder/pull/111 would fix the validation.

hfiguiere:
is this issue still a problem? Or can we just close it now.

PR #24, opened 2022-10-07
Add org.freedesktop.fwupd appstream check

hfiguiere:
See #23


hughsie:
If we merge https://github.com/flathub/org.flatpak.Builder/pull/111 then we don't need this exception.

PR #38, opened 2022-10-10
Add io.github.ezQuake exception

Eonfge:


Eonfge:
@hfiguiere Got you covered here

PR #40, opened 2022-10-11
Update org.telegram.desktop exceptions

gasinvein:


Erick555:
> "module-yasm-cmake-no-debuginfo": "Build-time only dependency, debuginfo is useless"

This is only warning. I don't think it makes sense to have exception for that. Moreover cmake inherits `-g` from `CFLAGS` so `-DCMAKE_BUILD_TYPE=Release` is useless for disabling debuginfo.

Erick555:
> "flathub-json-skip-appstream-check": "Upstream changes to metainfo, if fails validation, can block otherwise automatic updates"

Does it mean telegram is being published with invalid metainfo?

gasinvein:
> > "flathub-json-skip-appstream-check": "Upstream changes to metainfo, if fails validation, can block otherwise automatic updates"
> 
> Does it mean telegram is being published with invalid metainfo?

"Invalid" is a vast exaggeration. E.g. `appstream-cli` spits a warning on an URL in `<description>`, and treats warnings as validation failure by default. I do believe such issues are too minor to break the pipeline and delay updates.

Erick555:
Maybe appstream warnings shouldn't be treated as failures then? I don't see a sense for having strict checking then allow apps which don't pass to ignore it. What's the point? Either strict check should be mandatory or optional  for all.

hfiguiere:
we still use appstream-glib to validate AFAIK.

gasinvein:
> we still use appstream-glib to validate AFAIK.

It spits the same error:
```
builddir/files/share/appdata/org.telegram.desktop.appdata.xml: FAILED:
• style-invalid         : <release> description should be prose and not contain hyperlinks [<ul><li>A new expandable reaction menu was added in private chats and groups.</li><li>All users get access to dozens of new reactions, including reactions previously reserved for Premium users.</li><li>The reactions you use most frequently will always be displayed at the top.</li><li>Premium users can react to messages with thousands of custom emoji and add up to 3 reactions to each message.</li><li>Group admins can control whether custom reactions are allowed in their groups.</li><li>Premium users can add an animated emoji status next to their name to display their current activity.</li><li>To set a status, tap your Premium badge in the main menu or Settings.</li><li>Popular suggestions for working, sleeping, traveling and more will be shown at the top.</li><li>To set a status for a specific duration like 1 hour or 2 days, right click the emoji.</li><li>A new format was supported for username links, in addition to "t.me/username." You can now open Telegram accounts, groups or channels using links like "username.t.me" or "https://username.t.me."</li><li>Improved support for long media captions.</li></ul>]
• style-invalid         : Too many <p> tags for a good description [12/10]
Validation of files failed
```

Note that the metainfo releases are generated from official changelog that we aren't allowed to change and thus we can't just patch that.

PR #50, opened 2022-10-15
modules: change the CMake build type test

### LINKED ISSUE #41: cmake-no-debuginfo is bogus warning
Right now linter [checks](https://github.com/flathub/flatpak-builder-lint/blob/7db4a8b9b265fc3ac87f9980a571965f6b87f965/flatpak_builder_lint/checks/modules.py#L67) for `-DCMAKE_BUILD_TYPE=Release` and warns about supposedly missing debuginfo. This is misleading since cmake inherits `-g` from default `CFLAGS` environment which means `Release` has same amount of debuginfo as `RelWithDebInfo` has.

The actual difference is in optimization level `-O3` vs `-O2` which is different matter.

### LINKED ISSUE #47: module-{name}-cmake-non-release-build has too much false-positives
It doesn't treat MinSizeRel and random values as release builds. The latter is actively used to prevent falling back to a project-defined default CMAKE_BUILD_TYPE and don't have additional C(XX)FLAGS set by cmake. The right way is check whether the value is Debug, that's also how most cmake projects check to insert their own debug flags.

hfiguiere:
- it's a warning only
- list some know release

Issues:
  https://github.com/flathub/flatpak-builder-lint/issues/47 
  https://github.com/flathub/flatpak-builder-lint/issues/41

Should close #41 close #47

### DIFF
--- a/flatpak_builder_lint/checks/modules.py
+++ b/flatpak_builder_lint/checks/modules.py
@@ -64,11 +64,13 @@ class ModuleCheck(Check):
                         self.warnings.add(f"module-{name}-cmake-redundant-prefix")
                     elif opt.startswith("-DCMAKE_BUILD_TYPE"):
                         split = opt.split("=")
-                        if split[1] == "Release":
-                            # we shouldn't emit this on an extension
-                            self.warnings.add(f"module-{name}-cmake-no-debuginfo")
-                        elif split[1] != "RelWithDebInfo":
-                            self.errors.add(f"module-{name}-cmake-non-release-build")
+                        # There is too many possible choices and customizations.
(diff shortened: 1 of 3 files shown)


tsdgeos:
Trying to sanitize CMAKE_BUILD_TYPE makes no sense since it literally is a free text form

hfiguiere:
This PR make it just a warning though.

tsdgeos:
I understand, but warning about something that you are not sure it's wrong feels strange to me. But I'm just a random passer by, feel free to ignore :)

hfiguiere:
I don't exclude removing the warning in the future if the complaints outweight the benefit.

PR #52, opened 2022-10-15
Add exception for Flatseal

### LINKED ISSUE #51: Flatseal requires xdg-data access
See https://github.com/flathub/com.github.tchx84.Flatseal/pull/46
See https://github.com/tchx84/Flatseal/blob/master/com.github.tchx84.Flatseal.json#L14-L15

tchx84:
Closes #51


nanonyme:
It is good this is flagged though flagging could be more explicit. Any app with this permission can arbitrarily unsandbox itself similar to one with write access to home.

PR #61, opened 2022-10-20
Remove flathub-json-eol-message-without-rebase error

JakobDev:
A App can just not be developed anymore and is therefore EOL without a rebase


barthalion:
90% of time it's a mistake. Such apps should be submitted as exceptions instead of removing the check altogether.

PR #67, opened 2022-10-25
Added io.github.hakandundar34coding.system-monitoring-center into exceptions.json file

hakandundar34coding:


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1456,4 +1456,7 @@
     "org.srb2.SRB2Kart": {
         "flathub-json-modified-publish-delay": "extra-data"
     }
+    "io.github.hakandundar34coding.system-monitoring-center": {
+        "finish-args-flatpak-spawn-access": "the app predates this linter rule"
```
barthalion, line 1460: I mean, that's not true? Can you put actual use-case here?

hakandundar34coding: Updated the file again. A lot of applications used the previous explanation.

hfiguiere: because we had no explanation and we added the linter rules. This app does't predate the rule.

razzeee:
This was missing a leading `,` and this our json is not a valid json anymore :/

barthalion:
I'm aware, and fixing it right now.

hfiguiere:
Fixed with #73 I just merged. I also opened #74 to safeguard from this.

PR #71, opened 2022-10-26
Let skip --enable-debug=no

chmouel:
We are erroring out when we find a --enable-debug but we don't check if it was a actually an explicit =no to actually avoid the debug symbols.

Signed-off-by: Chmouel Boudjnah <chmouel@chmouel.com>


hfiguiere:
actually the debug symbols you want them they get split out into the .Debug. `--disable-debug` doesn't work? Debug shouldn't be the default though.

hfiguiere:
Merging this as this avoid a false positive.

PR #73, opened 2022-10-29
Fix exceptions json not being valid json

razzeee:


hfiguiere:
thanks!

hfiguiere:
We should add a validation of this file in CI

hfiguiere:
and github silently rebased this to inlude... nothing.

PR #79, opened 2022-11-04
exceptions: allow Fightcade to modify publishing delay

Pobega:
The Fightcade client (distributed on Flathub) must be in sync with the server version to connect to the service.

See https://github.com/flathub/flatpak-builder-lint/issues/28#issuecomment-1302811364 for more details.


Pobega:
(I don't really know what the deal is with the newline at the end of the file, I tried making the change both in Vim locally and via Github's web editor and it resulted in the same strange diff)

PR #80, opened 2022-11-09
update exceptions for com.gitlab.newsflash

jangernert:
It seems like `appid-code-hosting-too-few-components` was recently upgraded to throw an error. And now my flathub build failed:

https://github.com/flathub/com.gitlab.newsflash/pull/16

Do I have to wait for this to hit flathub to publish new releases?


barthalion:
No, exceptions are updated "separately".

PR #86, opened 2022-11-17
Allow finish-args-unnecessary-xdg-data-access for jdMinecraftLauncher

JakobDev:
jdMinecraftLauncher needs this to create menu shortcuts in share/applications.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -817,7 +817,8 @@
         "appid-uses-code-hosting-domain": "app-id predates this linter rule"
     },
     "com.gitlab.JakobDev.jdMinecraftLauncher": {
-        "appid-uses-code-hosting-domain": "app-id predates this linter rule"
+        "appid-uses-code-hosting-domain": "app-id predates this linter rule",
+        "finish-args-unnecessary-xdg-data-access": "jdMinecraftLauncher creates .desktop files in share/applications"
     },
     "com.gitlab.jakobdev.jdnbtexplorer": {
         "appid-uses-code-hosting-domain": "app-id predates this linter rule"


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -817,7 +817,8 @@
         "appid-uses-code-hosting-domain": "app-id predates this linter rule"
     },
     "com.gitlab.JakobDev.jdMinecraftLauncher": {
-        "appid-uses-code-hosting-domain": "app-id predates this linter rule"
+        "appid-uses-code-hosting-domain": "app-id predates this linter rule",
+        "finish-args-unnecessary-xdg-data-access": "jdMinecraftLauncher creates .desktop files in share/applications"
```
barthalion, line 821: Pretty sure this was added after jdMinecraftLauncher has been published to Flathub. It should be switched to use the dynamic portal that e.g. Epiphany uses to install PWAs like native apps.

JakobDev: > Pretty sure this was added after jdMinecraftLauncher has been published to Flathub.

That's right. the feature to create shortcuts have been added a few months after to the Program.

> It should be switched to use the dynamic portal that e.g. Epiphany uses to install PWAs like native apps.

If you are talking about the [DynamicLauncher](https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.DynamicLauncher) Portal, tthere are plenty of reasons to not use it.

1. It's pretty new, so not all Distros support it.

2. [It doesn't support creating files on the Desktop](https://github.com/flatpak/xdg-desktop-portal/issues/827), so I have to manage 2 different functions: One for the Portal and one for normally writing files on the Desktop.

3. [It doesn't work outside the sandbox](https://github.com/flatpak/xdg-desktop-portal/issues/826), so I still have to manage 2 different functions: One from inside the Flatpak and one for outside.

4. Even if all the problems above solved, I can't implement it. There is a Bug in pyQt6 that prevents this. I [reported a Bug](https://www.riverbankcomputing.com/pipermail/pyqt/2022-June/044757.html), but the answer was, that it could not been reproduced (even if another User could reproduce it), so this Bug is still not fixed today.

PR #89, opened 2022-11-20
Allow exceptions for de.schmidhuberj.tubefeeder

Schmiddiii:
These exceptions are needed as:

- `finish-args-flatpak-spawn-access`: Spawn any arbitrary video players for videos. `xdg-open` will not work as these videos are passed per a URL, `xdg-open` would therefore always open the browser.
- `finish-args-unnecessary-xdg-data-access`: Data about subscriptions, watchlist, filters is stored there. This was done before that rule existed.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1486,5 +1486,9 @@
     },
     "io.github.fsobolev.TimeSwitch": {
         "finish-args-flatpak-spawn-access": "required to execute user-defined commands on the host"
+    },
+    "de.schmidhuberj.tubefeeder": {
+        "finish-args-flatpak-spawn-access": "Spawn arbitrary video player",
+        "finish-args-unnecessary-xdg-data-access": "the app predates this linter rule"
     }
 }


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1486,5 +1486,9 @@
     },
     "io.github.fsobolev.TimeSwitch": {
         "finish-args-flatpak-spawn-access": "required to execute user-defined commands on the host"
+    },
+    "de.schmidhuberj.tubefeeder": {
+        "finish-args-flatpak-spawn-access": "Spawn arbitrary video player",
+        "finish-args-unnecessary-xdg-data-access": "the app predates this linter rule"
```
barthalion, line 1492: You will need to explain why your app needs direct access to xdg-config/mpv while it doesn't use mpv any more, and it uses xdg-cache/tubefeeder xdg-data/tubefeeder instead of respecting $XDG_CONFIG_HOME and $XDG_DATA_HOME, to which it has access by default

Schmiddiii: `xdg-config/mpv` is indeed not needed anymore. `xdg-cache/tubefeeder` is also fine inside the flatpak, is therefore also not needed anymore. With `xdg-data/tubefeeder` I have tried without and it does not read data from `~/.local/share/tubefeeder` (which is the directory I am using for legacy-reasons). I am using `glib::user_data_dir()` inside the code, so without would probably also work but would invalidate all the users data.

PR #97, opened 2022-12-01
Add install instructions to Readme

JakobDev:


bbhtt:
This is now documented here https://github.com/flathub/flatpak-builder-lint#local-environment, poetry is the better approach even though `pip` works.

PR #106, opened 2022-12-26
Allow EMPTY finish args

hfiguiere:
Valid case:

https://github.com/flathub/io.github.erkin.ponysay/pull/3#issuecomment-1365352224


hfiguiere:
better test to confirm the distinction between empty and missing.

barthalion:
Realistically, how often is this going to be legit?

hfiguiere:
I linked one case. But this patch address the empty vs missing. Empty show intent. Missing is still an error.

PR #99, opened 2022-12-06
Ignore command errors for BaseApp

JakobDev:

PR #111, opened 2023-01-07
BaseApp don't need a command

hfiguiere:


JakobDev:
I had made #99 which does the same but without the tests.  It's open for over one month without a response and I currently need it. I think it's time to ping @barthalion.

PR #113, opened 2023-01-16
exception: Add an exception for Kile

aleixpol:
It also needs shell access as it integrates a terminal.

@travier


aleixpol:
See https://github.com/flathub/org.kde.kile/pull/117

travier (approved):

PR #114, opened 2023-01-17
Update KDE tray permission error

### LINKED ISSUE #66: finish-args-broken-kde-tray-permission is incorrect
The `finish-args-broken-kde-tray-permission` forbids anything with `org.kde.StatusNotifierItem` at the beginning, however this isn't correct.

As you can see in knotification: https://github.com/KDE/knotifications/blob/7fb8c5b3130646845efb0483fc1cf3c7769c5830/src/kstatusnotifieritemdbus_p.cpp#L134

This is a unchanging format. In flatpak it will always be the same value for the same pid for the same item number. 

Now this is *very* broken inside of flatpak but it is the correct permission to say `--own-name=org.kde.StatusNotifierItem-2-1` because you know your values will always be the same. Fixing it means patching KDE libraries but that's a discussion for other people.

TingPing:
As of Qt 6.2 the tray works as expected without extra permissions. This version should be in all supported KDE runtimes.

Using org.kde.* is not a correct permission and is a security risk.

Closes #66

### DIFF
--- a/flatpak_builder_lint/checks/finish_args.py
+++ b/flatpak_builder_lint/checks/finish_args.py
@@ -53,7 +53,10 @@ class FinishArgsCheck(Check):
             self.errors.add("finish-args-redundant-home-and-host")
 
         for own_name in fa["own-name"]:
-            if own_name.startswith("org.kde.StatusNotifierItem"):
+            if (
+                own_name.startswith("org.kde.StatusNotifierItem")
+                or own_name == "org.kde.*"
+            ):
                 self.errors.add("finish-args-broken-kde-tray-permission")
 
             if appid:


barthalion:
We cannot introduce a breaking change like this without at least opening an issue against affected apps.

bbhtt:
> We cannot introduce a breaking change like this without at least opening an issue against affected apps.

I'm opening issues/PRs to convert apps from using `org.kde.*` when available.

bbhtt:
Can this be landed? I think I dealt with most of them, only 6 of my PRs are open right now https://github.com/search?q=author%3Abbhtt+is%3Apr+%22drop+org.kde%22+org%3Aflathub&type=pullrequests&p=1, most being merged.

Initially only, Discord, Discord Canary and Dropbox will need an exception.

Erick555:
The check was [dropped](https://github.com/flathub/flatpak-builder-lint/commit/aae87815476d5797f42dcf5d876c4568b024a9e7) completely (personally I would make it as warning).

bbhtt:
That was for `org.kde.StatusNotifierItem`, `org.kde.*` should be error-ed by default which this PR is about.

barthalion:
Yeah, `own_name == "org.kde.*"` should become an error once PRs are merged and exceptions in-place.

Erick555:
So this will break discord, zoom & co which is then reason for not making it.

bbhtt:
Probably add exemptions to them.

bbhtt:
Ok, I went through this again

> Lot of [apps](https://github.com/search?q=org%3Aflathub%20%22--own-name%3Dorg.kde.*%22&type=code) still use it

Only two are actionable (I'll open PR or issue) https://github.com/flathub/com.hunterwittenborn.Celeste and https://github.com/flathub/org.tribler.Tribler

The rest either all have open and working PRs, PRs but I didn't found a way to test, issues because they use old Electron/Qt etc. or are EOL and archived.

bbhtt:
This is continued in https://github.com/flathub-infra/flatpak-builder-lint/pull/259 and https://github.com/flathub-infra/flatpak-builder-lint/pull/262

PR #117, opened 2023-01-27
Downgrade KDE status icon error to a warning

foresto:
Remedies the build bot breakage described in #66, which is currently an incentive to grant dangerously broad dbus permissions, and currently prevents that dangerous permission from being fixed in cases where it was deployed in the past.


barthalion:
This is obviously not happening until the discussion at https://github.com/flathub/flatpak-builder-lint/issues/66 concludes.

TingPing:
I think this is OK for now. It will be months before the fixes land in Electron/Chromium.

PR #119, opened 2023-02-03
Add exemption for io.github.betaflight.BetaflightConfigurator

razzeee:


razzeee:
https://buildbot.flathub.org/#/builders/26/builds/2451

barthalion:
It shouldn't need access to the entire system bus. Did you try to nail down specific names with `--log-system-bus`?

hfiguiere:
`    "--system-talk-name=org.freedesktop.UPower"` doesn't work?

razzeee:
> ` "--system-talk-name=org.freedesktop.UPower"` doesn't work?

It didn't, when I used flatseal, but let me do it again, as a sanity check.

razzeee:
Nevermind, seems like I confused `talk` with `system-talk`...

PR #127, opened 2023-03-22
exceptions: Add font-viewer

A6GibKm:
The app requires the --filesystem=xdg-data/fonts:create permission to store/save fonts. This is a genuine usecase.


hfiguiere:
also predate the linter rule.

PR #133, opened 2023-04-09
Add "finish-args-flatpak-spawn-access" exception for com.dec05eba.gpu_screen_recorder

dec05eba:
The app requires flatpak-spawn to capture a monitor on AMD/Intel GPUs. It runs flatpak-spawn pkexec to run an external program as root (the program is part of the flatpak), because AMD/Intel requires root access to capture a monitor (KMS).

This is done to restrict root access to a very small executable that only does one thing (access to KMS), as opposed to running the whole flatpak (with network access) as root.


dec05eba:
> Could you resolve conflicts? Merged some other PRs and it broke this one in the process.

done

PR #139, opened 2023-05-13
Add appid-uses-code-hosting-domain exception for IP Lookup

Bytezz:
Needed for [com.github.bytezz.IPLookup](https://github.com/flathub/flathub/pull/4149)


Bytezz:
Since no response for a while, I converted id to io.github.

PR #140, opened 2023-05-17
Add org.fkoehler.KTailctl exception for access to flatpak-spawn.

f-koehler:
`org.fkoehler.KTailctl` fundamentally requires to run the Tailscale CLI tool in order to obtain information about the Tailscale status as well as controlling certain aspects of the client, e.g. connect/disconnect, sending/receiving files.


f-koehler:
Might not be necessary (see https://github.com/flathub/flathub/pull/4156). I will close this PR if integrating tailscale into the flatpak works.

f-koehler:
Integrating tailscale worked, so this change is not necessary.

PR #144, opened 2023-06-20
Allow Organic Maps beta channel to track latest upstream

bam80:
We pretend our master is good enough for testing by end users, so we would like to be able track upstream for flathub-beta channel.

Currently have error if commit is not mentioned:
https://buildbot.flathub.org/#/builders/26/builds/6014

This is the similar situation as https://github.com/flathub/flathub/issues/3588

@biodranik @barthalion

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -896,6 +896,9 @@
     "org.gnome.Totem.Devel": {
         "module-totem-source-git-no-commit-or-tag": "This is the development version of Totem, tracking upstream"
     },
+    "app.organicmaps.desktop": {
+        "module-organicmaps-source-git-no-commit-or-tag": "This is the beta version of Organic Maps, tracking upstream"
+    },
     "org.fcitx.Fcitx5": {
         "finish-args-arbitrary-dbus-access": "IDEs tend to require direct dbus access"
     },


hfiguiere:
flathub beta is NOT for nightly builds either. Build MUST be reproducible which mean bare branches are not allowed.

bam80:
But that's about exception, right? And it was accepted for Totem. We are at the same position I guess.

barthalion:
See, that's the problem with making exceptions. When you agree to do something once, it's only the matter of time before someone else reaches out. I'd rather rescind the Totem's exception that allow a new one.

barthalion:
It's not a big deal to update the hash of a commit to be built; it's no more work than having to go to buildbot and starting a new build. I will be contacting Totem maintainer about it.

bam80:
Do you know if we have similar concept as Edge channels on Snaps? Or maybe it was requested somewhere?

hfiguiere:
the two examples you give are operated by for profit companies.... (one being being worth TRILLIONS). Flathub is still run on a shoe string budget, and mostly operated by volunteers.

barthalion:
We're slowly opening up to builds uploaded from external CI systems, but even with that, disk space and bandwidth don't come for free (and I don't mean money here, we don't want regular use cases to be bottlenecked by nightly builds).

As opposed to Snap, anyone can host a Flatpak repo.

biodranik:
We publish releases from the master branch and try to keep it always releasable. It would be great to automate testing as there is a plan to make a great desktop/mobile Linux UI.

bam80:
> disk space and bandwidth don't come for free 

FWIW, `ostree` stores deltas only so it should be negligible?

nanonyme:
@bam80 no, deltas are mostly for in-transit optimization. ostree does store all data (deduplicated) normally.

bam80:
@nanonyme I'm not an expert but I'm not sure ostree stores the data deduplicated, at least until you didn't install both (mostly duplicated) flatpaks. Could you elaborate?

nanonyme:
@bam80 ostree stores same file exactly once assuming you handle metadata correctly. This allows you to have files stored only once if you have gnome, freedesktop-sdk and kde runtimes installed with matching base runtime.

nanonyme:
The thing is, there is such a concept in ostree as deltas. They are a delivery optimization method. They are not used for storage optimization.

bam80:
So if you have two _mostly_ the same files, they occupy twice storage in the repo on server anyway?

nanonyme:
Yes. You need to reproducibly produce exactly same files for dedplication.

bam80:
OK, I didn't know it (however, it's strange. I hope you know what you are saying).
But if you changed a few lines of code, the reproducibility won't help here, as the binary will be duplicated and it's clear now why this request can't be satisfied.
Probably we can close it then, thanks.

bbhtt:
> Probably we can close it then, thanks.

Closing as per above.

PR #150, opened 2023-07-18
Add exception for Turtlico

bbhtt:


hfiguiere:
Original submission was already a rename https://github.com/flathub/flathub/pull/1604

PR #152, opened 2023-07-30
Permit finish-args-flatpak-spawn-access for emacs

muep:


muep:
Based on the suggestion at https://github.com/flathub/org.gnu.emacs/pull/50 and some searching around, I get the impression that this would help move the aforementioned PR forward.

PR #159, opened 2023-08-14
Add flatpak-spawn exemption for Gradience

orowith2os:
Also possibly necessary for https://github.com/GradienceTeam/Gradience/issues/805, needed for https://github.com/flathub/com.github.GradienceTeam.Gradience/pull/13


orowith2os:
List of specific reasons for why Gradience needs this:

- Needs to install gnome-shell themes, and set GTK CSS files
- Needs to set global Flatpak permissions to allow :ro access to those GTK CSS files
- Needs to install the adw-gtk3 Flatpak theme
- Needs to configure dconf so the shell reads the gnome-shell theme
- Needs to set dconf so the GTK3 theme can be set by Gradience (TODO feature)
- Needs to read the host for all of the currently available themes

PR #168, opened 2023-09-11
Added the exceptions for my app.

SStoklosa23:
I added the finish-args-flatpak-spawn-access as I need it to launch the gnome-terminal from the app. My app uses it to connect to the ssh server the user wants.


hfiguiere:
what do you do if gnome terminal is not installed?

SStoklosa23:
The program will show an alert that says that the user does not have gnome-terminal installed.

hfiguiere:
I don't see the point of the flatpak if it require installing something like that.

SStoklosa23:
If gnome-terminal fails, it can check for the xdg terminal.

SStoklosa23:
try running x-terminal-emulator and you will see a new terminal.

SStoklosa23:
@hfiguiere

PR #169, opened 2023-09-17
Add spawn exception for turtle

philippun1:
Turtle is a development tool and needs to spawn a diff tool to compare files.


hfiguiere:
can't you ship the diff tool instead?

philippun1:
Interesting idea, I will try to do that first.

philippun1:
Adding it directly worked.

PR #172, opened 2023-09-22
Add io.github.MakovWait.Godots exception

CharlieQLe:


CharlieQLe:
It occurs to me that this doesn't actually have a description! This PR is necessary so any downloaded Godot editors within the flatpak can call host programs such as Blender, Visual Studio Code, etc., as Godot has integration with them whether it is a model importer or an IDE. This is for my PR [here](https://github.com/flathub/flathub/pull/4533).

PR #175, opened 2023-09-26
Add io.github.flattool.Warehouse

heliguy4599:


heliguy4599:
For clarity, Warehouse needs spawn permissions to manage flatpaks on the host (the main purpose of the app is to manage installed flatpaks and their user data)

PR #190, opened 2023-10-11
Add exceptions for org.godotengine.GodotSharp and org.godotengine.Godot3Sharp to allow both to use external code editors

Zishan-Rahman:
This is for these PRs in [flathub/flathub](https://github.com/flathub/flathub/), to allow both apps to use external editors using access to `flatpak-spawn`: https://github.com/flathub/flathub/pull/4594 and https://github.com/flathub/flathub/pull/4595

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1204,6 +1204,12 @@
     "org.godotengine.Godot3": {
         "finish-args-flatpak-spawn-access": "the app requires flatpak-spawn access to allow external code editors to be used with it"
     },
+    "org.godotengine.Godot3Sharp": {
+        "finish-args-flatpak-spawn-access": "the app requires flatpak-spawn access to allow external code editors to be used with it"
+    },
+    "org.godotengine.GodotSharp": {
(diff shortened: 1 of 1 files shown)


Zishan-Rahman:
> Let's avoid adding Godot 3 variant for now -- I'm not convinced it's worth the overhead to maintain 2 additional variants for apps already on Flathub. Can you change your PR to include only GodotSharp?

While I personally disagree with your point about the maintenance of both C# versions (I'd be happy with maintaining both concurrently), I _am_ OK with holding the Godot 3 variant off for a short while, as I'm currently trying to investigate [some problems I've been having with building it locally on my computer](https://github.com/flathub/flathub/pull/4595#discussion_r1358136416). I _have_ considered trying to do a test build of the Godot 3 variant without `flatpak-spawn` access just to see if those problems persist on a server build attempt rather than a local one, but if you're OK with that I'll do it once the Godot 4 variant's up on Flathub, since _that_ seems to build just fine. Once the Godot 4 variant's been published, I'd like to hopefully come back to the Godot 3 variant afterwards. I'm amending my PR now. Sorry for the late reply, by the way.

Zishan-Rahman:
> Can you change your PR to include only GodotSharp?

@barthalion (sorry for the ping) Just did :+1:

PR #197, opened 2023-10-29
Add flatpak-spawn exception for com.obsproject.Studio

GeorgesStavracas:
Until we reimplement virtual camera using PipeWire, this exception is needed.


tytan652:
Nitpik on the commit description

> Until we reimplement virtual camera using PipeWire, this exception is
needed.

Once implemented, the V4L2 implementation will be kept in deprecation for a given period of time.
We are going to need to keep the permission until the old implementation it's dropped, the actual description gives the wrong message.

PR #198, opened 2023-10-29
schema: Add 'secret-env' and 'secret-args'

GeorgesStavracas:
These are defined in flatpak-builder's schema as well.

Closes: https://github.com/flathub/flatpak-builder-lint/issues/196


tytan652:
You should make your changes by copy-pasting the schema from flatpak/flatpak-builder main branach or relying on https://github.com/flatpak/flatpak-builder/pull/543 and https://github.com/flatpak/flatpak-builder/pull/545.

PR #205, opened 2023-11-13
Add exception for gg.minion.Minion

zastrixarundell:
[Minion](https://github.com/flathub/gg.minion.Minion/) is an application / manager for AddOns in Elder Scrolls Online (ESO) and World of Warcraft (WoW).

It is the primary manager for ESO on all platforms. 

The default location for Steam compadata is within `xdg-data/Steam` and Minions needs to access it.

It can access `~/.local/share/Steam` without this PR but `xdg-data/Steam` would be better practice.

This has 3.5k downloads already <strike>, is in the process of being maintained by the actual owner and has been up for half a year already</strike>.

image

Here's a quick explanation on Linux about the application and small usage/setup example where it can bee seen that `xdg-data/Steam` could be accessed on a steamdeck (or native steam installation): https://www.youtube.com/watch?v=Xka0UalSjY0

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1449,6 +1449,9 @@
     "gg.guilded.Guilded": {
         "flathub-json-modified-publish-delay": "extra-data"
     },
+    "gg.minion.Minion": {
+        "finish-args-unnecessary-xdg-data-access": "This application manages AddOns in folders of game launchers and/or the folders within wine prefixes."
+    },
     "il.co.ravkavonline.RavKavOnline": {
         "flathub-json-modified-publish-delay": "extra-data"
     },


zastrixarundell:
This also does not expect these folders to be present, nor will it create them so `:create` is not needed. It only needs to access and modify the content inside them only if present, otherwise the user messed something up or is a non-standard installation of prefix locations.

zastrixarundell:
Update: the app became verified on flathub if that helps with this PR.

bbhtt:
Does the app follow `$XDG_DATA_HOME`? 

The difference between the two is `--filesystem=~/.local/share/steam` gives a more direct access while `xdg-data/Steam` gives it access and mounts it to `~/.var/app/<appid>/data` = (`$XDG_DATA_HOME`). So if the app is writing to `$XDG_DATA_HOME/Steam` it would work.

If it ignores `$XDG_DATA_HOME` and writes directly to `~/.local/share/steam` it should not make any difference.

zastrixarundell:
@bbhtt thanks for the input, I'm fairly new to XDG stuff myself.

It doesn't follow anything automatically, I the user needs to manually input the folder locations. I've just set these up as they're the default for the common used launchers. I've just understood that `xdg-data/Steam` is better practice than `~/.local/share/Steam`.

bbhtt:
The check got reverted in https://github.com/flathub/flatpak-builder-lint/pull/207, restart your builds after a few hours and it should work.

zastrixarundell:
Fixed due to #207.

PR #206, opened 2023-11-14
Add xdg-config for org.gimp.GIMP.

Jehan:
This permission was initially added for our use case (or at least partly, maybe others were asking too).
See: https://github.com/flatpak/flatpak/issues/328#issuecomment-260565160

We are actually considering dropping this for GIMP 3. But for the 2.10 series and the 2.99 development series, we need to keep continuity. We cannot suddenly break everyone's configuration in the middle of a release series.

Thanks!

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1512,6 +1512,9 @@
     "org.freefilesync.FreeFileSync": {
         "flathub-json-modified-publish-delay": "extra-data"
     },
+    "org.gimp.GIMP": {
+        "finish-args-unnecessary-xdg-config-access": "The option was created for us originally. See #328"
+    },
     "org.geogebra.GeoGebra": {
         "flathub-json-modified-publish-delay": "extra-data"
     },


bbhtt:
You also have a `xdg-config/GIMP` in the manifest. Isn't that supposed to follow `$XDG_CONFIG_HOME`?

Otherwise, seems fine.

hfiguiere:
the change in the linter might be reverted, it get triggered in conditions that never existed before.

Jehan:
> You also have a xdg-config/GIMP in the manifest. Isn't that supposed to follow $XDG_CONFIG_HOME?

Yeah why is that is all explained in discussions of flatpak/flatpak#328 though I summarized why we have this permission in this comment as @hfiguiere asked the same thing:
https://github.com/flathub/org.gimp.GIMP/pull/262#issuecomment-1810273394

hfiguiere:
Reverted with https://github.com/flathub/flatpak-builder-lint/pull/207

hfiguiere:
This is no longer needed. The build passed this stage.

PR #215, opened 2023-11-28
Allow ref_branch and target_repo mismatches for known branches

bbhtt:
```
Allow ref_branch and target_repo mismatches for known branches

Only verify that "stable" refs don't end up in beta repos and
vice-versa. For any other arbitrary ref branches it is not possible to
tell the intended target repo. So unless the ref_branch matches some
known branches that are already in use and have some significance return
"flat-manager-branch-repo" mismatch error
```

I assume the target_repo on flathub is fixed and can only be `stable`, `beta` or `test`.

The only other case where "branch" is used in manifest is for extensions, but those aren't checked here, and baseapps are exempt too. I could find only these two in known branches. If the list gets too long, might need to figure out a way to load it better.

I have no way of testing this due to not having access to a flathub like backend, hence draft.

### DIFF
--- a/flatpak_builder_lint/checks/flatmanager.py
+++ b/flatpak_builder_lint/checks/flatmanager.py
@@ -56,6 +56,8 @@ class FlatManagerCheck(Check):
                 for build_ref in build_extended.get("build_refs", [])
             ]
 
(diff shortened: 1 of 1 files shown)


bbhtt:
Sure

PR #223, opened 2023-12-08
io.github.erkin.ponysay: terminal only, no finish-args

hfiguiere:


hfiguiere:
see https://github.com/flathub/io.github.erkin.ponysay/pull/3

bbhtt:
Updated the docs in https://github.com/flathub/documentation/pull/142

PR #226, opened 2023-12-12
Add finish-args-unnecessary-xdg-data-access to jdFlatpakSnapshot

JakobDev:
This is for some reason now needed for the `--filesystem=xdg-data/flatpak:ro` permission.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1602,7 +1602,8 @@
         "finish-args-flatpak-spawn-access": "required for preview in Gnome Software"
     },
     "page.codeberg.JakobDev.jdFlatpakSnapshot": {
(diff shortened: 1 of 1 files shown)


bbhtt:
In constants.py you are directly using ~/.local/share/flatpak so why do you need that mounted inside the sandbox with `xdg-data/flatpak`? Shouldn't `--filesystem=~/.local/share/flatpak:ro` be enough? You are also calling flatpak on host.

JakobDev:
> In constants.py you are directly using ~/.local/share/flatpak

Looks like I forgot to change it to use the XDG Standard. Thanks for finding it. I will change this in a future Update, so I will need the Permission anyway.

> You are also calling flatpak on host.

Yes, why is this a problem?

bbhtt:
> Yes, why is this a problem?

Not an issue, I was just confused.

PR #227, opened 2023-12-13
Add org.ultimatepp.TheIDE to the exception list (finish-args-flatpak-spawn-access)

klugier:
In order to make proper Flatpak for TheIDE there is a need to use "--talk-name=org.freedesktop.Flatpak" permision. More details why it is need can be found in my [comment](https://github.com/flathub/flathub/pull/4778#issuecomment-1854030106) in the PR to main flathub repository.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1257,6 +1257,9 @@
     "org.texstudio.TeXstudio": {
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
+    "org.ultimatepp.TheIDE": {
+        "finish-args-flatpak-spawn-access": "the app predates this linter rule"
+    },
     "org.vim.Vim": {
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1257,6 +1257,9 @@
     "org.texstudio.TeXstudio": {
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
+    "org.ultimatepp.TheIDE": {
+        "finish-args-flatpak-spawn-access": "the app predates this linter rule"
```
hfiguiere, line 1261: what? This is not true. The linter rule exist while the package doesn't yet.

klugier: @hfiguiere  Can you tell me what should I do?

hfiguiere: like for any other exception request to put the actual reason, don't you think?

klugier: Done!

klugier:
Hello @hfiguiere, Can you confirm that everything is fine with this PR? I did what you expected.

PR #228, opened 2023-12-14
Print appstream validation errors

bbhtt:
This is an attempt to hopefully make the error a bit more readable, than printing the multiline output directly.

```
{
    "details": [
        "attribute-missing     : <release> has no timestamp",
        "tag-missing           : <content_rating> required [use https://odrs.gnome.org/oars]"
    ]
}
```
stderr is fail-safe, for invalid XMLs which stops at first failure eg. `"failed to parse com.example.app.metainfo.xml: Error on line 31 char 13: Element “component” was closed, but the currently open element is “summary”` (compose from flatpak-builder will fail before linter here)

I assume the first line will always be the filename/filepath from validation, so I strip that because its position gets messed up.

### DIFF
--- a/flatpak_builder_lint/checks/__init__.py
+++ b/flatpak_builder_lint/checks/__init__.py
@@ -18,6 +18,7 @@ class Check(metaclass=CheckMeta):
     warnings: Set[Optional[str]] = set()
     errors: Set[Optional[str]] = set()
     jsonschema: Set[Optional[str]] = set()
+    appstream: Set[Optional[str]] = set()
     repo_primary_ref: Optional[str] = None
 
     def _populate_ref(self, repo: str) -> None:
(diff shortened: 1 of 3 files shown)


hfiguiere:
this is probably a good step.

hfiguiere:
> stderr is fail-safe, for invalid XMLs which stops at first failure eg. `"failed to parse com.example.app.metainfo.xml: Error on line 31 char 13: Element “component” was closed, but the currently open element is “summary”` (compose from flatpak-builder will fail before linter here)

This is highly dependent on the output of the cli. And might break anytime but as we somewhat control it, maybe it's ok. Can still be fixed.

hfiguiere (approved):
LGTM

barthalion:
Either `details` should be renamed to `appstream` to match existing `jsoncheck`, or should be reworked to place either into a matching subkey.

bbhtt:
I'll rename it, I don't like putting them into other ones.

hfiguiere (approved):
We should try to land this ASAP. The failures are counter-productive at that point.

PR #229, opened 2023-12-14
Update exceptions.json adding RetroDECK xdg-access

XargonWan:
From https://github.com/flathub/flathub/issues/4786

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1148,7 +1148,7 @@
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
     "net.retrodeck.retrodeck": {
-        "finish-args-unnecessary-xdg-data-access": "SteamOS is the primary target of the application and it does not ship xdg-desktop-portal-gtk: https://github.com/flathub/flathub/issues/4786"
+        "finish-args-unnecessary-xdg-data-access": "SteamOS is the primary target of the application and it does not ship an xsettings daemon or xdg-desktop-portal-gtk: https://github.com/ValveSoftware/SteamOS/issues/803"
     },
     "net.veloren.veloren": {
         "flathub-json-modified-publish-delay": "Continuous builds automated and synced with upstream CI. Server and client must use identical build."
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1147,6 +1147,10 @@
     "net.lutris.Lutris": {
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
+    "net.retrodeck.retrodeck": {
+        "appid-uses-code-hosting-domain": "app-id predates this linter rule",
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1147,6 +1147,10 @@
     "net.lutris.Lutris": {
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
+    "net.retrodeck.retrodeck": {
+        "appid-uses-code-hosting-domain": "app-id predates this linter rule",
```
bbhtt, line 1151: net.retrodeck doesn't seem a code hosting domain?

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1147,6 +1147,10 @@
...
+        "finish-args-unnecessary-xdg-data-access": "https://github.com/flathub/flathub/issues/4786",
```
bbhtt, line 1152: The current usage is for gtk-3.0 folder, why is that needed? Also, can you please put the actual reason here, not the link.

flatpak_builder_lint/staticfiles/exceptions.json
hfiguiere: also don't lie. this doesn't predate a linter rule, because there is none against that app-id.

hfiguiere:
Closing as per matrix conversation and https://github.com/flathub/flatpak-builder-lint/pull/229#discussion_r1426685382

XargonWan:
Turns out that that arg is needed for displaying the dark theme on GTK.
It seems to be a Valve side bug that is under review: https://github.com/ValveSoftware/SteamOS/issues/803
SteamOS 3 doesn't have GTK portal.

flatpak_builder_lint/staticfiles/exceptions.json
hfiguiere: this need to be resolved, it's not an option

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1147,6 +1147,10 @@
     "net.lutris.Lutris": {
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
+    "net.retrodeck.retrodeck": {
+        "appid-uses-code-hosting-domain": "app-id predates this linter rule",
+        "finish-args-unnecessary-xdg-data-access": "SteamOS is the primary target of the application and it does not ship xdg-desktop-portal-gtk: https://github.com/flathub/flathub/issues/4786"
```
TingPing, line 1152:
```suggestion
        "finish-args-unnecessary-xdg-data-access": "SteamOS is the primary target of the application and it does not ship an xsettings daemon or xdg-desktop-portal-gtk: https://github.com/ValveSoftware/SteamOS/issues/803"
```

bbhtt (approved):
LGTM, because the app primarily targets steam deck which lacks necessary desktop integration to load GTK themes.

PR #231, opened 2023-12-14
Allow finish-args-unnecessary-xdg-data-access for Gear Lever

mijorus:


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1622,6 +1622,7 @@
         "finish-args-unnecessary-xdg-data-access": "jdMinecraftLauncher creates .desktop files in share/applications"
     },
     "it.mijorus.gearlever": {
+	"finish-args-unnecessary-xdg-data-access": "Required to add desktop files to the app menu",
```
bbhtt, line 1625: The indentation is wrong

bbhtt:
You are accessing the path relative to home, github.com/mijorus/gearlever/blob/90583c27ad5e18a6545442f5f7d41137d4f26f05/src/providers/AppImageProvider.py#L76 not `XDG_DATA_DIR`, so why use xdg-data? You can use `filesystem=~/.local/share/applications` which isn't gated by linter.

Also, https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.DynamicLauncher.html allows installing desktop files without any holes.

mijorus:
Also, I guess filesystem=home should be enough.
Thanks, I'm closing this

PR #233, opened 2023-12-15
Add xdg-data exception for com.calibre_ebook.calibre

Erick555:
This is required for trash access because app doesn't use trash portal.

See https://github.com/flathub/com.calibre_ebook.calibre/issues/83


bbhtt (approved):
LGTM, it tries to access trash directory relative to `$XDG_DATA_HOME` https://github.com/kovidgoyal/calibre/blob/eb78a761a99ac20a6364f85e12059fec6517d890/src/calibre/utils/linux_trash.py#L164-L169

PR #234, opened 2023-12-15
Add xdg-data exception for page.codeberg.libre_menu_editor.LibreMenuEditor

libre-menu-editor:
This program is a Menu Editor. It needs write access to xdg-data/applications and read access to xdg-data/flatpak.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -7,6 +7,9 @@
     "org.flathub.exceptions_wildcard": {
         "*": "ignore all errors and warnings"
     },
+    "page.codeberg.libre_menu_editor.LibreMenuEditor": {
+        "finish-args-unnecessary-xdg-data-access": "app-id predates this linter rule"
+    },
     "com.github.sixpounder.GameOfLife": {
         "appid-uses-code-hosting-domain": "app-id predates this linter rule"
     },
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -8,7 +8,7 @@
         "*": "ignore all errors and warnings"
     },
     "page.codeberg.libre_menu_editor.LibreMenuEditor": {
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -7,6 +7,9 @@
     "org.flathub.exceptions_wildcard": {
         "*": "ignore all errors and warnings"
     },
+    "page.codeberg.libre_menu_editor.LibreMenuEditor": {
+        "finish-args-unnecessary-xdg-data-access": "app-id predates this linter rule"
```
hfiguiere, line 11: please write an actually reason instead if senselessly cut and pasting incorrect information.

libre-menu-editor: This information is not incorrect. I wrote and submitted this app before these new linter-rules existed. I just want to push an update to my app

libre-menu-editor:
I provided a different reason. Hopefully, this one will be accepted

bbhtt:
Why do you need `xdg-data/flatpak` and not `xdg-data/flatpak/exports/share/applications`? Seems like this is a menu editor.

libre-menu-editor:
> Why do you need `xdg-data/flatpak` and not `xdg-data/flatpak/exports/share/applications`? Seems like this is a menu editor.

This will not be enough if symlinks are involved. I also need access to the location the symlinks are pointing to.

I had this conversation already when submitting the app. Please just add the exception. I prioritize functionality over minimal permissions and do not want end-users to have to deal with errors.

Manifest: https://github.com/flathub/page.codeberg.libre_menu_editor.LibreMenuEditor/blob/master/page.codeberg.libre_menu_editor.LibreMenuEditor.yml

bbhtt:
I didn't know that, I wasn't around when the initial submission happened. But seems ok, as there is no portal solution to this and the app is accessing the paths relative to [XDG_DATA_DIR correctly](https://codeberg.org/libre-menu-editor/libre-menu-editor/src/commit/c5f129ab8164a950c97c3a0ac8efd37c0f38b647/libre-menu-editor/modules/gui.py)

bbhtt (approved):

PR #236, opened 2023-12-18
Add exception for xdg-config for org.ryujinx.Ryujinx

marysaka:
This is used for GTK3 theme access.


bbhtt:
Reading gtk-3.0/settings.ini for the theme is wrong. If on X11 the theme is communicated to sandbox applications via xsettings daemon, if on Wayland gnome, it is communicated via xdg-portal-gnome. On KDE you need to explicitly set gtk theme in system settings.

tldr, theme should be communicated to apps without any permissions if the desktop has proper xsettings daemon/portal running.

marysaka:
Well I removed that for org.ryujinx.Ryujinx so I guess no need to add it back.

Thank you for the info :+1:

PR #240, opened 2023-12-22
exceptions: org.kde.gwenview: Add finish-args-unnecessary-xdg-data-access

travier:
See: https://github.com/flathub/org.kde.gwenview/pull/117


travier:
Let's add an exception while we sort out if Gwenview needs still needs Trash access and if portal support needs to be added.

bbhtt:
lgtm, looks like kio::trash is being used, which searches/creates the trash folder relative to XDG_DATA_DIR on Linux. I didn't see KIO's linux trash implementation supporting the portal.

bbhtt (approved):

Issue #156, opened 2023-08-08
Add basic appdata quality checks


bertob:
We have new [appdata quality guidelines](https://docs.flathub.org/docs/for-app-authors/appdata-guidelines/quality-guidelines) now, some of which could be relatively easily checked in an automated way, including:

- [ ] Icon fills the entire canvas (no transparent pixels)
- [ ] Icon is a symbolic (i.e. all non-transparent pixels are the same color)
- [x] Summary ending with dot (".")
- [x] Name >20 characters
- [x] Summary >35 characters
- [x] No screenshot captions
- [x] Developer field is not empty

hfiguiere:
That require writing a brand new tool that will be run on the content of the flatpak AFTER the build.

razzeee:
@bertob we should probably mention, that we only want to look at the english strings for now. (I have been thinking about translation in the past days)

bbhtt:
3, 4 and 5 are in https://github.com/flathub-infra/flatpak-builder-lint/pull/246. Neither appstream-util or appstreamcli warns on 3 and 4. appstreamcli warns for 5 only in pedantic mode.

bbhtt:
The icon checks are in the website moderation board nowadays. I don't think adding them in the linter makes sense, because at most we can raise a warning which is not very useful.

The moderation dashboard can provide a UI for the canvas etc.

PR #246, opened 2024-01-01
Warn on too long names, summary, missing screenshot captions and project_license

bbhtt:
See https://github.com/flathub-infra/flatpak-builder-lint/issues/156

When name is > 20 characters or summary is > 35 characters

appstreamcli does not warn this, even in pedantic mode.


bbhtt:
I plan to reuse appstream.py for metainfo, screenshots and desktop file. Otherwise it's the same thing being copied too many times.

`types-lxml` is to provide type hints for lxml, hopefully it works. mypy had to be bumped for that and this bug https://github.com/python/mypy/issues/13627 which prevented using types-lxml.

bbhtt:
Hmm, I just realised names and summary can both have `xml:lang` although it doesn't seem to mentioned here https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-name I need to go through all of them instead of the first one.

bbhtt:
Added a warning on missing `project_license`, appstream-glib allows missing it. There are too many apps missing it sometimes from major desktop groups, so can't do an error but it should be an error otherwise licensing becomes dubious.

hfiguiere:
I do think we should move towards making this an error.

bbhtt:
I would like some data on Flathub before maing it an error.

I'm not against making it an error.

razzeee:
Wondering if we might want to stick these into something like 'appstream-quality-...'

bbhtt:
I don't have a preference.

barthalion:
LGTM. Do we know how many apps will end up failing this?

barthalion:
Welp, there's a merge conflict now. Mind rebasing?

bbhtt:
> LGTM. Do we know how many apps will end up failing this?

Probably a lot, but no idea. I made all the checks warnings for now to stay on the safe side\.

razzeee:
> LGTM. Do we know how many apps will end up failing this?

Login at flathub.org and go to the statistics page

bbhtt:
The quality guidelines are specific to Flathub, appstreamcli is meant to be used outside of it too. The spec doesn't mandate a length

razzeee:
The lengths should also always just be warnings

barthalion:
Warnings should be considered "future deprecations", at least in the current form. People generally don't act upon messages unless they cause CI failures.

PR #247, opened 2024-01-04
Add a check for low quality categories

bbhtt:
See https://github.com/ximion/appstream/issues/577

`Categories=GTK;Qt;` will be caught but not `Categories=GTK;Qt;Network;` because there was at least one item not in `block`.

### DIFF
--- a/flatpak_builder_lint/checks/desktop.py
+++ b/flatpak_builder_lint/checks/desktop.py
@@ -1,7 +1,7 @@
 import os
 import subprocess
 import tempfile
-from typing import Optional
+from typing import Optional, Set
 
 from lxml import etree  # type: ignore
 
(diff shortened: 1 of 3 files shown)


flatpak_builder_lint/checks/desktop.py
```diff
@@ -116,6 +116,19 @@ def _validate(self, path: str, appid: str) -> None:
...
+            try:
+                # https://github.com/ximion/appstream/blob/146099484012397f166cd428c56f230487b2d1fc/src/as-desktop-entry.c#L144-L154
+                # appstreamcli filters these out during compose, "GUI"
+                # is caught by desktop-file-validate
+                block = {"KDE", "GTK", "Qt", "Application", "GNOME"}
+                catgr: Set[str] = set(
+                    filter(None, d["Desktop Entry"]["Categories"].split(";"))
+                )
+                if len(catgr) > 0 and catgr.issubset(block):
+                    self.warnings.add("desktop-file-low-quality-category")
```
bbhtt, line 128: Should this be a warning or an error? I kept it warning to stay safe.

barthalion: Let's stick to a warning until we settle on a sustainable way of communicating these.

PR #249, opened 2024-01-04
metainfo: Drop flathub.json handling and rework

bbhtt:
`flathub.json` has never worked unless it was installed in builddir. 

I don't see a reason to keep it and complicate things, considering metainfo checks can be skipped based on the heuristics alone and anything not skipped ends up being a few regular desktop-applications.

Now they should be checked, but `flathub.json` gives them a silent way out of the linter's metainfo checks without being gated by exceptions.

Any legitimate desktop-applications needing to use `skip-appstream-check` or `skip-icon-check` should apply for the corresponding linter exception. I can probably ease some of it by mass adding some legitimate exceptions.

The state with this PR:

- Extensions/baseapps skips all checks here early
- No check is done unless `type=desktop, desktop-application, console-application`
- Icon check is done only on `type==desktop,desktop-application`
- Rest ie. `desktop, desktop-application, console-application` goes through appinfo + metainfo checks and validation

Draft because I need to test it locally + discussions.

I can probably create a version of this PR with `flathub_json` included if need be.

### DIFF
--- a/flatpak_builder_lint/checks/metainfo.py
+++ b/flatpak_builder_lint/checks/metainfo.py
@@ -15,7 +15,6 @@ class MetainfoCheck(Check):
             f"{path}/files/share/appdata",
         ]
         metainfo_exts = [".appdata.xml", ".metainfo.xml"]
-        metainfo_path = None
 
         if appid.endswith(".BaseApp"):
             return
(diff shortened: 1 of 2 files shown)


bbhtt:
Marking ready because after some local testing it seems to be working fine.

PR #257, opened 2024-01-13
Add xdg-data exception to PikaBackup

sophie-h:


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1718,6 +1718,9 @@
     "app.drey.KeyRack": {
         "finish-args-unnecessary-xdg-data-access": "the app predates this linter rule"
     },
+    "org.gnome.World.PikaBackup": {
+        "finish-args-unnecessary-xdg-data-access": "the app predates this linter rule"
```
bbhtt, line 1722: Please put an actual reason. The predate ones were auto imported when the linter was first introduced, manual ones should not be using that reason.

sophie-h: Same MR for different app was merged 7 days before. https://github.com/flathub-infra/flatpak-builder-lint/pull/255

Anyways, I updated the reason.

Erick555: You could use `--filesystem=~/.local/share/flatpak` instead and it would work without. override

sophie-h: Does that mean that Flatpak hard-codes the path instead of using the XDG Base Directory spec?

bbhtt: No, if your application follows `$XDG_DATA_HOME`, the permission is correct and the exception is needed.

If it follows `$HOME`, that one would work.

It's also the reason, why a proper reasoning is needed. A lot of applications copy-paste this permission without actually needing it.

bbhtt: See https://docs.flatpak.org/en/latest/sandbox-permissions-reference.html#f6

bbhtt: I'm confused... You are already using ` --filesystem=xdg-data/flatpak:ro` in the manifest.

I'm saying that is the correct permission to use...

Is that not working?

sophie-h: It was working until it was blocked by the linter. I just tried to explain that this is not about the app following the xdg spec but rather flatpak itself by overmounting the dir with a tmpdir that we mount again.

Erick555: > That means your solution would for example cause overrides to not be included in the backup if the XDG_DATA_HOME is different. Or am I missing something?

Yes, in case `XDG_DATA_HOME` is non empty o host then it wouldn't work. It's pretty extreme case which would break dozens apps in flathub which hardcode this path though.

sophie-h: Why do we have linters if the suggested solution is to bypass them by using a hard-coded path to do the same instead?

PR #258, opened 2024-01-14
Command line package exceptions for se.emijoh.mpw

emiljoha:
Based on other exceptions for these liner I get the impression they are not applicable to terminal application like `se.emijoh.mpw`

https://github.com/flathub/se.emijoh.mpw/pull/6

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1190,6 +1190,10 @@
     "org.freedesktop.appstream-glib": {
         "flathub-json-skip-appstream-check": "It is a command line package"
     },
+    "se.emijoh.mpw": {
+        "flathub-json-skip-appstream-check": "It is a command line package",
+	"finish-args-not-defined": "It is a command line package"
+    },
     "org.freedesktop.LinuxAudio.BaseExtension": {
         "toplevel-no-command": "It is a base app without anything to run",
         "finish-args-not-defined": "It is a base app without anything to run",
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1191,7 +1191,6 @@
         "flathub-json-skip-appstream-check": "It is a command line package"
     },
     "se.emijoh.mpw": {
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1190,6 +1190,10 @@
     "org.freedesktop.appstream-glib": {
         "flathub-json-skip-appstream-check": "It is a command line package"
     },
+    "se.emijoh.mpw": {
+        "flathub-json-skip-appstream-check": "It is a command line package",
+	"finish-args-not-defined": "It is a command line package"
```
bbhtt, line 1195: Please fix the indentation here

hfiguiere:
There is used to be a trick to bypass the need for an exception, based on python parser behaviour of the manfiest, but it stopped working when the linter started checking the flatpak metadata file.

So yeah it's necessary.

hfiguiere:
as for the appstream-check, please don't make it validate.

Notably GPL-3.0 is an invalid metadata license (that's the appstream file itself). Add screenshots and release. This is regardless of being a CLI

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1190,6 +1190,10 @@
     "org.freedesktop.appstream-glib": {
         "flathub-json-skip-appstream-check": "It is a command line package"
     },
+    "se.emijoh.mpw": {
+        "flathub-json-skip-appstream-check": "It is a command line package",
```
hfiguiere, line 1194: no. a CLI app should have a valid appstream file.

flatpak_builder_lint/staticfiles/exceptions.json
emiljoha: Spaces instead of tabs? 0c024a9

flatpak_builder_lint/staticfiles/exceptions.json
emiljoha: Will fix the appstream file instead! 630719e

emiljoha:
@bbhtt @hfiguiere  Thank you so much for the fast review :heart:

bbhtt (approved):

Issue #178, opened 2023-10-01
Investigate if `org.freedesktop.portal.*` permissions are needed


bbhtt:
Afaiu, these are allowed by flatpak itself https://github.com/flatpak/flatpak/blob/f66b1ecb7420ba56bc57cada92a1ed01a5a330f9/common/flatpak-run-dbus.c#L290-L294 and there should be no reason to put a `--talk-name` to them in app manifest since that allows for wider access than what flatpak gives by default.

Around ~36 apps are using it based on a Github search

The most common one being `org.freedesktop.portal.Fcitx` which seems to have [originated](https://github.com/flatpak/flatpak/issues/2031#issuecomment-482567827) from a fcitx bug back in 2018 https://github.com/flatpak/flatpak/issues/2031 and since been [fixed](https://github.com/fcitx/fcitx/commit/1f4c3a96cb2d1308581864676dd86ea6aa393ccb) on fcitx 4.2.9.7 (even Debian oldstable has newer fcitx than that https://packages.debian.org/search?keywords=fcitx&searchon=names&suite=all&section=all)

Another one popular one seems to be the background portal which seems to be not required at all to keep the app alive eg. https://github.com/flathub/org.signal.Signal/pull/524

I think these should be error-ed by default in the linter

bbhtt:
Hi @smcv, can you please comment here, if this is correct?

ebassi:
Correct: interfaces under `org.freedesktop.portal.` are automatically added to the allowlist, otherwise portals would never work out of the box.

bbhtt:
Thanks!

Just for completeness, I asked in #xdg-desktop-portals:matrix.org to confirm my understanding in OP before I went on to break a bunch of app builds.

PR #259, opened 2024-01-15
Block wildcard access of bus names in well-known namespace

bbhtt:
Alternative to https://github.com/flathub-infra/flatpak-builder-lint/pull/114, closes https://github.com/flathub-infra/flatpak-builder-lint/pull/114, blocks in both `own-name` and `talk-name`.

List based on ones that are most likely to find their way in manifests and can potentially give widespread access. If there are more, please comment below.

This can be reused to block the pattern in https://github.com/flathub-infra/flatpak-builder-lint/issues/178 too

I don't see a generic method to block `*` because a bunch of legitimate ones can end in `*` The third component or ones after that are reserved for applications, e.g. `org.gnome.evolution.*`

Closes https://github.com/flathub-infra/flatpak-builder-lint/issues/178

### DIFF
--- a/flatpak_builder_lint/checks/finish_args.py
+++ b/flatpak_builder_lint/checks/finish_args.py
@@ -69,6 +69,20 @@ class FinishArgsCheck(Check):
                     own_name.startswith(appid) and own_name[len(appid)] == "."
                 ):
                     self.errors.add("finish-args-unnecessary-appid-own-name")
+            if own_name == "org.freedesktop.*":
+                self.errors.add("finish-args-wildcard-freedesktop-own-name")
+            if own_name == "org.gnome.*":
+                self.errors.add("finish-args-wildcard-gnome-own-name")
+            if own_name == "org.kde.*":
+                self.errors.add("finish-args-wildcard-kde-own-name")
+
+        for talk_name in finish_args["talk-name"]:
+            if talk_name == "org.freedesktop.*":
+                self.errors.add("finish-args-wildcard-freedesktop-talk-name")
+            if talk_name == "org.gnome.*":
+                self.errors.add("finish-args-wildcard-gnome-talk-name")
+            if talk_name == "org.kde.*":
+                self.errors.add("finish-args-wildcard-kde-talk-name")
 
         if (
             "xdg-config/autostart" in finish_args["filesystem"]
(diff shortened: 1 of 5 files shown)


bbhtt:
I'll import an exception list (mainly for `org.kde.*`) if this is fine.

barthalion:
Yeah, go ahead

bbhtt:
I thought about it again a bit and having separate errors would be better because having a combined exception allows an app to have all 3 bus access when it say needs only "org.kde.*" for legacy purposes.

I'll raise a separate error for each.

bbhtt:
Just for more fine-grained exception system...

flatpak_builder_lint/checks/finish_args.py
```diff
@@ -69,6 +69,24 @@ def _validate(self, appid: Optional[str], finish_args: dict[str, Set[str]]) -> N
...
+                self.errors.add("finish-args-portal-own-name")
+
+        for talk_name in finish_args["talk-name"]:
+            if talk_name == "org.freedesktop.*":
+                self.errors.add("finish-args-wildcard-freedesktop-talk-name")
+            if talk_name == "org.gnome.*":
+                self.errors.add("finish-args-wildcard-gnome-talk-name")
+            if talk_name == "org.kde.*":
+                self.errors.add("finish-args-wildcard-kde-talk-name")
+            if talk_name.startswith("org.freedesktop.portal"):
```
hfiguiere, line 88:
```suggestion
            if talk_name.startswith("org.freedesktop.portal."):
```

I'd add the period here. and also above.

bbhtt:
> maybe we should start by submitting PR....

I'll open some issues later.

bbhtt:
Opening of issues is done, I count 31 cross-links in https://github.com/flathub-infra/flatpak-builder-lint/issues/178

bbhtt:
`org.kde.*`:

Rest have issues open and are cross-linked to https://github.com/flathub-infra/flatpak-builder-lint/pull/114.

All except Quaternion and Chiaki are unmaintained/had little to no activity. The maintainer of Chiaki is not sure why the permission is in the manifest but did not want to remove the permission. So that one is unsolvable from my side.

`org.freedesktop.portal.*`:

Waiting for a bit so that more apps can respond to the issues I opened.

PR #261, opened 2024-01-17
Validate desktop-files of console-applications too

bbhtt:
Console applications can supply desktop files too

Console applications are allowed to have:

(1): No desktop file present
(2): No Icon key in desktop file
(3): NoDisplay=True
(4): No icons installed

But if they install a desktop file, it must:

(1): Pass validation using desktop-file-validate
(2): Not have empty or a wrong value in Icon key
(3): Have a proper Exec key
(4): Not use Hidden=True [1]
(5): Use Terminal=True if NoDisplay is unset or set to False ie. the
desktop file is meant to be displayed on the desktop environment.

[1]: Hidden is equivalent to the desktop file not existing at all per the
spec [2]. If console apps want to hide the desktop file from the desktop
they should be using `NoDisplay=true`. The error is renamed to
separate this use-case and allow for exceptions.

[2]: https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys


hfiguiere:
did I miss where it checks `Terminal=true` ?

bbhtt:
It doesn't. I wasn't sure if every console application should have/supposed to have "Terminal=true"

flatpak_builder_lint/checks/desktop.py
```diff
@@ -99,19 +109,21 @@ def _validate(self, path: str, appid: str) -> None:
...
-            try:
-                nodisplay = d["Desktop Entry"]["NoDisplay"]
-                if nodisplay == "true":
-                    self.errors.add("desktop-file-is-hidden")
-            except KeyError:
-                pass
+            if not is_console:
+
+                try:
+                    hidden = d["Desktop Entry"]["Hidden"]
```
hfiguiere, line 115: this is should stay even for console applications.

> Hidden should have been called Deleted. It means the user deleted (at their level) something that was present (at an upper level, e.g. in the system dirs). It's strictly equivalent to the .desktop file not existing at all, as far as that user is concerned. This can also be used to "uninstall" existing files (e.g. due to a renaming) - by letting make install install a file with Hidden=true in it.

bbhtt: Hmm yea, I'll change it

bbhtt: It's so used interchangeably with NoDisplay that a bunch of console applications will probably fail.

bbhtt:
> It doesn't. I wasn't sure if every console application should have/supposed to have "Terminal=true"

Now it does

bbhtt:
rebased over the glib.keyfile port.

tests/test_builddir.py
```diff
@@ -108,16 +108,26 @@ def test_builddir_extension() -> None:
...
     ret = run_checks("tests/builddir/console")
     found_errors = set(ret["errors"])
 
+    assert "appstream-unsupported-component-type" not in found_errors
     assert errors == found_errors
 
 
+def test_builddir_appstream_unsupported_ctype() -> None:
+
+    ret = run_checks("tests/builddir/appstream-unsupported-ctype")
```
bbhtt, line 125: Forgot to commit this, I'll push it later.

PR #267, opened 2024-01-19
desktop: Refactor to support checking multiple desktop files

bbhtt:


flatpak_builder_lint/checks/desktop.py
```diff
@@ -8,9 +9,36 @@
...
+                line = line.strip()
+                if line.startswith("#") or line == "" or line == "\n":
+                    continue
+                if line.startswith("[") and line.endswith("]"):
+                    group = line[1:-1]
+                    data[group] = {}
+                if "=" in line:
+                    key = line.split("=", 1)[0]
+                    value = line.split("=", 1)[1]
+                    data[group][key] = value
```
JakobDev, line 25: This is a bit self promotion, but you could take a look at [desktop-entry-lib](https://pypi.org/project/desktop-entry-lib/) which is already battle tested and covered with unit tests instead of writing a custom parser.

bbhtt: Yea sure I can look into it but not in this PR.

PR #268, opened 2024-01-24
Add xdg-data exception to Maniatic Launcher

santiagocezar:


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1677,6 +1677,9 @@
     "io.github.Youda008.DoomRunner": {
         "finish-args-flatpak-spawn-access": "Needed to execute external game engines"
     },
+    "io.github.santiagocezar.maniatic-launcher": {
+        "finish-args-unnecessary-xdg-data-access": "Needed to copy the assets from ~/.local/share/Steam/steamapps/common/Sonic Mania/Data.rsdk into the CWD"
```
bbhtt, line 1681: Looks like you are accessing the paths relative to home and not XDG_DATA_DIR https://github.com/santiagocezar/maniatic-launcher/blob/06bdf2233df4ddf1690e078b826d775fb7b82294/maniatic/setup_game.py#L8

So `--filesystem=~/.var/app/<steam app id>` or `~/.local/share/steam` or `~/.steam` should be enough, you should not need this.

santiagocezar: Ah! I thought that `~/.local/share/` would trigger the linter anyways. Thanks!

PR #269, opened 2024-01-24
Add check for uncessary gamescope permission

TingPing:
The correct way to use gamescope is simply to run:

    gamescope --expose-wayland flatpak run ...

### DIFF
--- a/flatpak_builder_lint/checks/finish_args.py
+++ b/flatpak_builder_lint/checks/finish_args.py
@@ -57,6 +57,8 @@ class FinishArgsCheck(Check):
                     self.errors.add(f"finish-args-reserved-{resv_dir}")
             if fs.startswith("/home") or fs.startswith("/var/home"):
                 self.errors.add("finish-args-absolute-home-path")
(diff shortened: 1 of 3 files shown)


TingPing:
Actually before this gets merged, this may not be possible to do when launching from the Steam UI...

TingPing:
So by default Steam of course cannot see any Flatpaks and it depends on the tool users use to add them to Steam.

In this one for example it could easily set `export WAYLAND_DISPLAY=gamescope-0`:

https://github.com/dragoonDorise/EmuDeck/blob/main/tools/launchers/dolphin-emu.sh

But we'd have to get involved with these communities to do this.

bbhtt:
Currently we can't really add this to the blocklist, because on the steam deck users can't easily modify the arguments passed to gamescope.

`xdg-run/gamescope-0` is just the default, it can be gamescope-1,2 ... so static permission is somewhat insufficient. The solution would be to they expose wayland by default https://github.com/ValveSoftware/gamescope/pull/1112

Closing.

PR #276, opened 2024-01-30
Add exception to io.kapsa.drive; Allow to setup FUSE network mount - S3Drive

tomekit:
Hi,

We would appreciate if you can review this exception for our app (https://flathub.org/apps/io.kapsa.drive).

One of our functionalities is to setup FUSE mount on the host system and we've got it running in this pull request: https://github.com/flathub/io.kapsa.drive/pull/30 however build fails (https://buildbot.flathub.org/#/builders/6/builds/96828) due to: 
```
{
    "errors": [
        "finish-args-flatpak-spawn-access"
    ]
}
```

hence our exception request.

Screenshot from 2024-01-30 23-37-24
Screenshot from 2024-01-30 23-37-44

Thanks !


barthalion:
Yeah, it may take up to an hour, depending on when a new publish job triggers a refresh.

Issue #83, opened 2022-02-22
Filter or reject AppStream metadata with invalid icons


smcv:
`cc.nift.nsm` currently has AppStream metadata with:

```
 <icon height="128" width="128">$APPID.png</icon> 
 <icon height="64" width="64">$APPID.png</icon> 
```

(https://github.com/flathub/cc.nift.nsm/issues/3)

This seems wrong in a couple of ways: libappstream warns on reading this (although apparently appstream-glib considers it to be perfectly valid?), and apps probably also shouldn't be allowed to specify an icon name that is outside their namespace.

The maintainer of libappstream [says](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1006269) this is something that Flatpak and/or Flathub should be preventing:

> There seems to be something wrong with Flatpak's or Flathub's
> AppStream generator (even if the input is bad, the output of the
> server shouldn't be bad too!)

(But I don't think validating AppStream is really Flatpak's or flatpak-builder's job, and it seems more appropriate for this to be something that Flathub is responsible for.)

barthalion:
Just did.

n-ham:
If anyone does know what the issue is can fix it up. I don't know
Flatpack/hub very well and have done what others have suggested in the past
but still seems to cause problems down the road..

razzeee:
Hrm, I thought you we're good now. And this issue now is about how to prevent this in the future.

smcv:
> Hrm, I thought you [were] good now. And this issue now is about how to prevent this in the future.

That matches my understanding: I think `cc.nift.nsm` was fixed a while ago and current versions are OK. The feature I'm requesting in this issue is that if someone makes the same mistake in future, then the Flathub infrastructure should not publish that package.

ximion:
The AppStream validator will catch this and complain, so if `validate` or `validate-tree` is used with `appstreamcli`, this wouldn't have made it (you can also make any issue fatal with `--override`, so some Flathub-specific stricter checks are possible).
Not sure if that is done right now though...

number3d:
it's been over a year. can we already throw out cc.nift.nsm completely? 

alternatively, is there a way to remove appstream and keep Gnome install happy on my arch?

PR #277, opened 2024-02-02
metainfo: Add a check for icon key and type

bbhtt:
For desktop applications. Shows error if there is no icon key present or icon key has no type or type remote.

Closes https://github.com/flathub-infra/flatpak-builder-lint/issues/83

My understanding is icons should always be of type stock, cached or local on Flathub; remote is disallowed and no type is just wrong


bbhtt:
Actually looks, there's currently no issue with `remote` icons?

Seems to be using it https://flathub.org/apps/app.moosync.moosync and is fine.

barthalion:
No, this is right. We want to block maintainers from using remote icons.

flatpak_builder_lint/checks/metainfo.py
```diff
@@ -67,6 +67,12 @@ def _validate(self, path: str, appid: str) -> None:
         ):
             if not os.path.exists(appinfo_icon_path):
                 self.errors.add("appstream-missing-icon-file")
+            if not appstream.has_icon_key(appstream_path):
+                self.errors.add("appstream-missing-icon-key")
```
barthalion, line 71: Just to make sure, we're running this only on desktop and console apps?

flatpak_builder_lint/checks/metainfo.py
```diff
@@ -67,6 +67,12 @@ def _validate(self, path: str, appid: str) -> None:
...
+            if appstream.is_icon_remote(appstream_path):
+                self.errors.add("appstream-remote-icon-key")
```
barthalion, line 73: Somewhat confusingly, there are apps mixing local and remote icons. We should not fail if both are defined.

flatpak_builder_lint/checks/metainfo.py
bbhtt: Only for desktop and desktop-application types in appstream. Some console apps ship with no icon key in metainfo or install an icon.

flatpak_builder_lint/checks/metainfo.py
bbhtt: Just to clarify, you mean `type=local`?

flatpak_builder_lint/appstream.py
```diff
@@ -74,3 +74,21 @@ def has_manifest_key(path: str) -> bool:
...
+
+
+def has_icon_key(path: str) -> bool:
+    return bool(components(path)[0].xpath("icon"))
+
+
+def is_icon_remote(path: str) -> bool:
+    for icon in parse_xml(path).findall("component/icon"):
+        if "remote" in str(icon.attrib):
+            return True
```
bbhtt, line 86: Needs to be modified to check that remote icons have texts starting with https://dl.flathub.org/media

If not it's considered broken.

flatpak_builder_lint/checks/metainfo.py
bbhtt: With the check modified this doesn't matter, it'll fail only when the remote icon entry doesn't start with Flathub url.

https://github.com/flathub-infra/flatpak-builder-lint/pull/277#discussion_r1490270136

PR #278, opened 2024-02-04
Add xdg-config exception of dev.k8slens.OpenLens

proletarius101:
Related to https://github.com/flathub/dev.k8slens.OpenLens/issues/146

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1073,6 +1073,9 @@
     "dev.salaniLeo.immagini": {
         "finish-args-flatpak-spawn-access": "needed to start applications"
     },
+    "dev.k8slens.OpenLens": {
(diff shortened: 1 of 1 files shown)


bbhtt:
Does it actually need `xdg-config` or can `~/.config/gcloud` work? I can't find where the path is defined, presumably some closed source google's component that the app is bundling?

bbhtt:
> I'm actually requesting for xdg-config/gcloud and I thought it was a more resilient way to write ~/.config/gcloud?

xdg-config will mount the directory to the `XDG_CONFIG_HOME` directory inside flatpak. If the program doesn't search relative to `XDG_CONFIG_HOME` but hardcodes `$HOME` for example, you don't need that and `~/.config/gcloud` is enough.

bbhtt:
I assume this won't be needed anymore. But feel free to reopen or ping me, if it is needed.

Issue #280, opened 2024-02-05
Flathub strips out captions from appstream data


bbhtt:
For some reason Flathub, server side is stripping out captions from appstream catalogue data. The same is not reproducible when building locally using the exact same commands and `org.flatpak.Builder`

The caption check should be re-enabled once the catalogue data has captions, otherwise the check has no meaning

See also https://github.com/flathub-infra/website/issues/48

bbhtt:
I'm disabling the check for now.

hadess:
Ran into this building https://github.com/flathub/com.github.shonumi.gbe-plus/, which would complain about `appstream-screenshot-missing-caption` even though I just [added captions](https://github.com/flathub/com.github.shonumi.gbe-plus/blob/master/com.github.shonumi.gbe-plus.appdata.xml#L22).

Locally, it doesn't complain about it:
```sh
$ flatpak run --command=flatpak-builder-lint org.flatpak.Builder builddir _build/
{
    "errors": [
        "appid-uses-code-hosting-domain"
    ],
    "warnings": [
        "finish-args-redundant-device-all"
    ]
}

Please consult the documentation at https://docs.flathub.org/docs/for-app-authors/linter
```

bbhtt:
Yea that's what confused me too. Clearly, something is going wrong on the server, but I don't have the power to debug that.

bbhtt:
Solved by porting to libappstream, will re-enable the check once more apps get rebuilt and composes their metadata with libappstream

```
cat appstream.xml|grep "<caption>"
        <caption>Screnshoot 1</caption>
        <caption>Classic View</caption>
        <caption>RHDC View</caption>
        <caption>Image of Arduino IDE v2</caption>
        <caption>Simple view</caption>
        <caption>Advanced view</caption>
        <caption>Recording page</caption>
        <caption>Talk on an invite only server</caption>
        <caption>Voice and Video</caption>
        <caption>Dark Mode Window</caption>
        <caption>Light Mode Window</caption>
        <caption>Available on Linux, macOS, Windows, iOS, Android, and your web browser</caption>
        <caption>Running the benchmark</caption>
        <caption>CLEAR AND SIMPLE UI</caption>
        <caption>PERFECT FOR ORIGINAL PRUSA</caption>
        <caption>BUILT-IN 3RD PARTY PRINTER PROFILES</caption>
        <caption>MSLA AND MULTI MATERIAL SUPPORT</caption>
[...]

PR #281, opened 2024-02-05
Disable caption check until catalogue data stops stripping it out

bbhtt:
See https://github.com/flathub-infra/flatpak-builder-lint/issues/280


wjt (approved):

wjt:
Lgtm (though I do not have commit access).

PR #294, opened 2024-02-17
Add a check to catch absolute /run/media paths

bbhtt:
Only `/run/media` and `/run/media/` are allowed. Any subpaths
makes it specific to how it is being mounted by host.


bbhtt:
@hfiguiere

PR #296, opened 2024-02-18
cli: Add a helper for running appstream validation

barthalion:


flatpak_builder_lint/cli.py
```diff
@@ -165,12 +165,18 @@ def main() -> int:
...
-    if results := run_checks(args.type, path, args.exceptions, args.appid):
-        if "errors" in results:
-            exit_code = 1
+    if args.type != "appstream":
+        if results := run_checks(args.type, path, args.exceptions, args.appid):
+            if "errors" in results:
+                exit_code = 1
 
         output = json.dumps(results, indent=4)
         print(output)
```
bbhtt, line 174: I think this needs to be moved a level deep under the second if condition.

Otherwise it'll print out an empty json on the other checks when there's nothing to print.

bbhtt:
Otherwise seems ok.

PR #298, opened 2024-02-18
Provide multi-arch container images

darkdragon-001:
This provides an arm64 image which is needed for https://github.com/flathub-infra/flatpak-github-actions/commit/79d4438081ca6fbab81225518b16a4887ae667e9 to work.


darkdragon-001:
O_o you build outside of the container and only copy the artifacts inside the container. Should be fixed in #300.

PR #300, opened 2024-02-18
Fix multi-arch container

darkdragon-001:
Move build steps into container to actually build for other architectures. Fixes arm64 support added in #298.


darkdragon-001:
@barthalion `pull_request_target` uses the workflow file from `main` which still has the step to build `flatpak-builder` outside of the `Dockerfile`. But as the new `Dockerfile` doesn't copy these files in, this could actually work.

darkdragon-001:
@barthalion Also as the event is not `pull_request` but `pull_request_target` this will actually push the image with tag `latest`.

darkdragon-001:
`bwrap: Creating new namespace failed: Operation not permitted`, I guess we need to run the container as `--privileged` or run `flatpak-builder` with `--disable-rofiles-fuse`?

darkdragon-001:
`flatpak-github-actions` uses `--privileged`: https://github.com/flatpak/flatpak-github-actions/blob/b8be6bc0ede43cb3a2cc55895654312b14f84a0d/README.md?plain=1#L31. I will try with `--disable-rofiles-fuse`.

darkdragon-001:
> They use it to _run_ the image, not build it.

But it is the image running `flatpak-builder`...

darkdragon-001:
Did I disable fuse correctly?
I get
```
#20 104.4 FB: Running: flatpak build --die-with-parent --env=FLATPAK_BUILDER_BUILDDIR=/run/build/libxcvt --nofilesystem=host:reset --filesystem=/build/org.flatpak.Builder/.flatpak-builder/build/libxcvt-1 --bind-mount=/run/build/libxcvt=/build/org.flatpak.Builder/.flatpak-builder/build/libxcvt-1 --build-dir=/run/build/libxcvt/_flatpak_build --bind-mount=/run/ccache=/build/org.flatpak.Builder/.flatpak-builder/ccache --env=SOURCE_DATE_EPOCH=1708294716 '--env=CFLAGS=-O2 -pipe -g -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer ' '--env=CXXFLAGS=-O2 -pipe -g -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer ' '--env=LDFLAGS=-L/app/lib -Wl,-z,relro,-z,now -Wl,--as-needed ' --env=V=1 --env=MOUNT_FUSE_PATH=../tmp/
#20 104.4 bwrap: Creating new namespace failed: Operation not permitted
```

darkdragon-001:
Seems to work with `security.insecure` which the BuildKit's equivalent to `--privileged` (see last commit).

darkdragon-001:
With `security.insecure`, I now get `Creating new namespace failed: Invalid argument`.

docker/build.sh
```diff
@@ -1,17 +1,28 @@
...
+        arch=aarch64
+        ;;
+esac
+
 cd org.flatpak.Builder 
-flatpak-builder --user --verbose --force-clean --repo=repo \
-    --ccache --install-deps-from=flathub \
-    builddir org.flatpak.Builder.json
-rm -rf files/lib/debug
+flatpak-builder --arch=$arch --user --verbose --force-clean --repo=repo \
```
darkdragon-001, line 26: How does it work? Does the manifest have a cross-compilation toolchain set up correctly? Or does it use QEMU as well to execute binaries from the SDK in the target architecture?

hfiguiere: There is no cross-compilation with flaptak (ie compiler running on a platform that is not the target). It will run everything in QEMU.

darkdragon-001:
It doesn't seem to be possible to create manifests for images which are not yet pushed. I created the steps as I think they work but can't test them without permissions to push to a registry. @barthalion could you please take over from here?

darkdragon-001:
@barthalion Thank you so much for your help! Works like a charm!

PR #301, opened 2024-02-19
Include org.gnome.FontManager exceptions

Eonfge:
See https://github.com/flathub/org.gnome.FontManager/issues/29

@barthalion Based on your comments


bbhtt:
CI failure is unrelated.

PR #310, opened 2024-02-23
Add finish-args-contains-both-x11-and-wayland to jdMinecraftLauncher

JakobDev:


bbhtt:
This is a warning, not an error. You don't need an exception.

bbhtt:
It was an error last year, but was downgraded because Thunderbird, Firefox lots of electron apps needed both.

Only recently some of them became native wayland and removed the permission.

PR #318, opened 2024-02-26
Port build metadata parser to GLib keyfile

bbhtt:


bbhtt:
I'm guessing the manifest here needs to be updated but I think the deps should already be in either via the Flatpak builder baseapp or runtime, I can check later

https://github.com/flathub-infra/flatpak-builder-lint/blob/master/docker/flatpak-builder-lint-deps.json

PR #328, opened 2024-02-29
exception: Added Inkscape xdg-config

hfiguiere:


bbhtt:
https://gitlab.com/inkscape/inbox/-/issues/2862

hfiguiere:
For the record it was set before the linter rule worked.

PR #331, opened 2024-03-03
add exception of org.DolphinEmu.dolphin-emu

cpba:
The id `org.DolphinEmu.dolphin-emu` fails validation because of the uppercase letters, however, I'm unable to change it.

Failed build:

https://buildbot.flathub.org/#/builders/6/builds/104677

Pull request I'm trying to merge:

https://github.com/flathub/org.DolphinEmu.dolphin-emu/pull/190

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -2411,5 +2411,8 @@
     },
     "org.linuxshowplayer.LinuxShowPlayer": {
         "finish-args-flatpak-spawn-access": "Required by CommandCue(s) to allow users to interact with external tools"
+    },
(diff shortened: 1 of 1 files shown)


cpba:
Currently the application doesn't work at all because of an issue caused by an abi change in the underlying runtime. It'd need a rebuild in order to work again.

In the case an exception can't be granted, would it be possible to at least trigger a rebuild so that users are able to use the app again?

Link to issue: https://github.com/flathub/org.DolphinEmu.dolphin-emu/issues/189

bbhtt:
Unfortunately, we can't grant you an exception for this. The repo name will be fixed by an admin to make it lowercase and an eol-rebase will be set up serverside.

bbhtt:
should be solved now

PR #333, opened 2024-03-04
finish-args: redundant-device-all should not trigger when shm is present

bbhtt:
all does not cover shm

man flatpak-metadata
> /dev/shm (which is separately specified)

Seperate the tests to that affect too


hfiguiere:
ideally no one needs `shm`. it was added with the belief it would solve JACK support, but then it turned out that JACK client and server (that use shm as a communication layer) must match. So it became pointless quickly.

hfiguiere:
LGTM though

bbhtt:
Yea still want to consider the case in code.

PR #335, opened 2024-03-09
Port desktop file parser to GLib.keyfile

bbhtt:


flatpak_builder_lint/checks/desktop.py
```diff
@@ -2,30 +2,14 @@
 import re
 import subprocess
 import tempfile
-from typing import Optional, Set
+
+from gi.repository import GLib
```
hfiguiere, line 6: do we need to add gi-python (or whatever it's called)?

It's only on the GNOME runtime.

bbhtt: PyGobject? It's in the baseapp already, we use it already for the flatpak metadata parser in linter and flat manager.

PR #336, opened 2024-03-10
Add flatpak-spawn exception for dev.neovide.Neovide

sqwxl:
This exception is needed to allow Neovide to execute the host's `nvim` binary.


bbhtt:
LGTM, essentially a neovim frontend, wants to access nvim and other tools from host. nvim has this already.

bbhtt (approved):

PR #337, opened 2024-03-10
Add xdg-cache exception to org.gnome.Contacts

A6GibKm:
Fixes https://github.com/flathub/org.gnome.Contacts/pull/84.

**cc** @nielsdg

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -892,7 +892,8 @@
         "finish-args-arbitrary-dbus-access": "the app requires direct dbus access"
     },
     "org.gnome.Contacts": {
-        "finish-args-unnecessary-xdg-data-access": "the app predates this linter rule"
+        "finish-args-unnecessary-xdg-data-access": "the app predates this linter rule",
+        "finish-args-unnecessary-xdg-cache-access": "the app predates this linter rule, used for evolution data server"
     },
     "org.gnome.dspy": {
         "finish-args-arbitrary-dbus-access": "the app requires direct dbus access"


bbhtt:
Something is likely wrong with `xdg-data/pixmaps` permission, contacts searcher for pixmaps under Environment.get_system_data_dirs() which gives it `/app/share:/usr/share:/usr/share/runtime/share:/run/host/user-share:/run/host/share` inside the sandbox, not sure why it is asking `xdg-data/pixmaps` since that is user_data_dir and not in the search path.

I see that permission in the manifest from the very [first commit](https://github.com/flathub/org.gnome.Contacts/commit/c7a8082576857ffb1823d7ff4f6e2f2a38b740b3) of contacts on Flathub, so something likely changed.

LGTM, eds runs on host and uses that as [storage path](https://github.com/GNOME/evolution-data-server/blob/ed2541eb77ff98105a72042ea2b4f05a107a24b7/src/addressbook/libedata-book/e-book-backend.c#L291-L308), contacts uses the same thing.

bbhtt (approved):

bbhtt:
oh woops the fixes thing closed the PR

A6GibKm:
> Something is likely wrong with `xdg-data/pixmaps` permission, contacts searcher for pixmaps under Environment.get_system_data_dirs() which gives it `/app/share:/usr/share:/usr/share/runtime/share:/run/host/user-share:/run/host/share` inside the sandbox, not sure why it is asking `xdg-data/pixmaps` since that is user_data_dir and not in the search path.
> 
> I see that permission in the manifest from the very [first commit](https://github.com/flathub/org.gnome.Contacts/commit/c7a8082576857ffb1823d7ff4f6e2f2a38b740b3) of contacts on Flathub, so something likely changed.
> 
> Otherwise, LGTM, eds runs on host and uses that as [storage path](https://github.com/GNOME/evolution-data-server/blob/ed2541eb77ff98105a72042ea2b4f05a107a24b7/src/addressbook/libedata-book/e-book-backend.c#L291-L308), contacts uses the same thing.

A question for @nielsdg, but there was a good reason for it. Thanks!

nielsdg:
Ehh, this is from a long time ago really :sweat_smile: 

I can think of 2 possible reasons:

One was that it was a very liberal implementation of the icon theme spec a bit (see https://specifications.freedesktop.org/icon-theme-spec/latest/ar01s03.html), allowing several of the listed directories to be combined:

>  By default, apps should look in $HOME/.icons (for backwards compatibility), in $XDG_DATA_DIRS/icons and in /usr/share/pixmaps (in that order)

The other possibility was that we just followed from wherever those avatars got installed. These days, I think it's coming from g-c-c, but in earlier times, IIRC we used to get it from somewhere else (although I don't recall the details of that anymore).

PR #344, opened 2024-03-15
exceptions.json: allow flatpak-spawn --host from gplaces

dimkr:
gplaces uses user-defined programs to handle non-text MIME types (for example, the default handler for images is `chafa`) and URL schemes it doesn't support (defaults to `xdg-open`). Without working `flatpak-spawn --host`, gplaces can be annoying to use because one needs to download a file and open it using a host application, or copy the URL and paste it in the browser.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -226,7 +226,8 @@
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
     "com.github.dimkr.gplaces": {
-        "appid-uses-code-hosting-domain": "app-id predates this linter rule"
+        "appid-uses-code-hosting-domain": "app-id predates this linter rule",
+        "finish-args-flatpak-spawn-access": "required to handle unsupported MIME types and URLs"
     },
     "com.github.donadigo.appeditor": {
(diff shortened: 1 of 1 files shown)


bbhtt:
If you use xdg-open, your desktop environment is supposed to show an app chooser dialogue with programs that can handle that mimetype.

bbhtt:
There is an entry for this app on line 228

dimkr:
> Why would you need the ability to launch them directly?

gplaces doesn't use mimeapps.list: it has a configuration file that defines the command-line to use, per MIME type or URL scheme. It runs these handler programs directly, so it can stream to stdin (for example, it can download a video and write it to stdin of mpv). In addition, the default applications is not guaranteed to be a sane default for gplaces: the default image viewer is likely to be a graphical application, while for gplaces, you'd probably want something like chafa.

(The handler for https:// URLs is xdg-open, in the default gplaces configuration file, is xdg-open)

bbhtt:
Do you have an example of a gemini site hosting images that I can test out the feature with? It seems very hard to find one and the mp3 example in your readme seems to time out.

bbhtt:
Thanks, should be up in 10-15 minutes.

bbhtt (approved):

dimkr:
Thank you!

Issue #345, opened 2024-03-21
Disallow branches altogether in sources


bbhtt:
Only branch: already blocked

branch + commit: If HEAD of branch does not point to the commit, build fails.

hfiguiere:
This happened today.

Linter would have prevented it. 

I remember clearly making this check fatal.

bbhtt:
Changed in https://github.com/flathub-infra/flatpak-builder-lint/commit/883cc94d42b616703373eec5c9409b2771d37a4d

I think we can safely disallow branch + commit combination after notifying apps using it

bbhtt:
Ok using a commit hash in branch actually works even when HEAD of branch changes, it behaves just like using only a commit.

There's no possible combination in which both branches and a commit is required. I think I can remove the `if not commit` part.

hfiguiere:
a plain commit has always worked.

I think I remember one issue where someone wanted to always update to the tip of the branch and f-e-d-c would use that. It's probably the only (albeit still broken IMHO) case. Instead f-e-d-c should just accept "branch" as a parameter.

hfiguiere:
What I see is not making it an error if there is a f-e-d-c stanza.

bbhtt:
One issue it doesn't require a fedc stanza to update them, fedc will auto update the commit to the latest one if `branch: branch_name` is present.

This is how the x-checker flatpak works, see for example https://github.com/flathub/org.flathub.flatpak-external-data-checker/pull/245/files

hfiguiere:
oh. TIL

bbhtt:
Non exhaustive list of apps this is going to break:

- [ ] https://github.com/flathub/com.github.Matoking.protontricks/blob/master/com.github.Matoking.protontricks.yml
- [ ] https://github.com/flathub/org.goldendict.GoldenDict/blob/master/org.goldendict.GoldenDict.yml
- [ ] https://github.com/flathub/org.musicbrainz.Picard/blob/master/org.musicbrainz.Picard.json
- [ ] https://github.com/flathub/rocks.koreader.KOReader/blob/master/rocks.koreader.KOReader.yml
- [ ] https://github.com/flathub/com.github.KRTirtho.Spotube/blob/master/com.github.KRTirtho.Spotube.yml
- [ ] https://github.com/flathub/io.github.JaGoLi.ytdl_gui/blob/master/io.github.JaGoLi.ytdl_gui.yaml
[...]

PR #346, opened 2024-03-21
modules: Disallow git branches in modules even if commit is present

bbhtt:
It is only allowed when the `branch` value is itself a commit hash.

Only `branch` with an actual branch name is non-reproducible, a `branch` and a `commit` means build fails whenever they mismatch.

Fix and seperate tests to that effect.

Closes https://github.com/flathub-infra/flatpak-builder-lint/issues/345

### DIFF
--- a/flatpak_builder_lint/checks/modules.py
+++ b/flatpak_builder_lint/checks/modules.py
@@ -23,12 +23,11 @@ class ModuleCheck(Check):
             tag = source.get("tag")
             branch_committish = False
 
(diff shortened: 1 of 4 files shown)


bbhtt:
This is a breaking change for manifests using `branch: branch_name` and `commit: commit_hash` we need to notify as such

bbhtt:
Let me know if there are some allowed/disallowed combinations I'm missing.

PR #348, opened 2024-03-27
appstream: Require that appstream IDs match $FLATPAK_ID

bbhtt:
For historical reasons flatpak-builder passes ``$FLATPAK_ID.desktop` as a valid component ID to appstreamcli. This mismatch creates issues like the one described in https://github.com/flatpak/flatpak/pull/5752 and https://github.com/flathub/flathub/issues/4222

So require that both match and avoid new submissions leaking in with a mismatch.


bbhtt:
This is a breaking change, we need to put exceptions for everything listed in https://github.com/flathub/flathub/issues/4222#issuecomment-2020140606

Things I didn't check for: 

- apps with appstream starting with `application` but still having mismatch are not in the list
- I put this above `desktop, desktop-application, console-application` to ensure nothing having a catalogue data gets through here

bbhtt:
Should be ready now

bbhtt:
I think you are at the wrong PR. I merged this linter check, the Flatpak PR didn't get reviews.

ximion:
Indeed - I had two tabs open and chose the wrong one to reply to somehow - oops :sweat_smile:

PR #166, opened 2023-09-04
chore/ci: bump action versions in workflows

kbdharun:
## Changes

- Bump `actions/checkout` from `v3` (eol Node 16) to `v4` (Node 20).
- Bump `actions/setup-python` from `v4` (eol Node 16) to `v5` (Node 20).

PR #361, opened 2024-04-06
chore: add dependabot configuration for scheduled dependency updates

kbdharun:
## Changes

Continuing #166, this PR adds a dependabot config file to schedule automatic dependency updates (with PRs by dependabot) in a monthly schedule. This would ensure the dependencies are up to date and would allow addressing any breaking changes in the dependent packages then and there.

## Detailed Explanation of changes

```md
version: 2 # this is the version type of the dependabot config file with 2 being the recent one
updates:
  - package-ecosystem: "github-actions" # checks for updates to the GitHub actions in use.
    directory: "/" # scans the full root directory for updates (optionally, we can make this fine-grained to point to .github/workflows)
    schedule:
      interval: "monthly" # scheduling dependabot updates monthly

  - package-ecosystem: "pip"  # checks for updates to the Python (Poetry) deps in use.
    directory: "/" # scans the full root directory for updates
    schedule:
      interval: "monthly" # scheduling dependabot updates monthly
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "monthly"


bbhtt:
Thanks for the PR. Dependabot is not super high on the list of things to add because the CI setup of linter depends on `org.flatpak.Builder` where the linter is actually shipped and the manifests there needs to be manually updated and kept in sync with the dependencies here. So just merging dependabot updates won't help.

PR #373, opened 2024-04-20
metainfo: Check if launchable defined in appstream is actually installed

bbhtt:
For legacy ids that end in `.desktop`, appstreamcli validate does not raise the usual launchable missing error here and assumes the file is present. compose also ignores that the file does not exist

### DIFF
--- a/flatpak_builder_lint/appstream.py
+++ b/flatpak_builder_lint/appstream.py
@@ -68,6 +68,11 @@ def appstream_id(path: str) -> Optional[str]:
     return str(aps_cid)
 
 
+def get_launchable(path: str) -> list:
+    launchable = components(path)[0].xpath("launchable[@type='desktop-id']/text()")
+    return list(launchable)
+
+
 def is_developer_name_present(path: str) -> bool:
     developer_name = components(path)[0].xpath("developer/name")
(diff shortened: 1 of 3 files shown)


bbhtt:
This shouldn't be a breaking change because we already had https://github.com/flathub-infra/flatpak-builder-lint/blob/027e4eeb9c1560acd773dfd9de60cb37c33452fb/flatpak_builder_lint/appstream.py#L27 for everything else and without the launchable tag compose would fail anyway on `icon-not-found, gui-app-no-icon, category-invalid`

bbhtt:
```python
from lxml import etree

# curl -sL https://hub.flathub.org/repo/appstream/x86_64/appstream.xml.gz | gunzip
components = etree.parse("appstream.xml").getroot()
for component in components.iterchildren(tag="component"):
   if component.attrib.get("type") not in ("desktop","desktop-application"):
        continue
   appstream_id = component.find("id").text
   launchable = component.xpath("launchable[@type='desktop-id']/text()")
   if not launchable:
      print(appstream_id)
```

shows 

```
com.lettier.movie-monad.desktop
org.sparkleshare.SparkleShare.desktop
```

PR #382, opened 2024-04-22
desktop: Check that icon file from desktop file is actually installed

bbhtt:
This should not be breaking as without the correct icon from the desktop file installed, appstream will not create `files/share/app-info/icons/{appid}.png`.

We already check that `files/share/app-info/icons/{appid}.png` exists and icons following the flatpak id patterns exist in `files/share/icons/hicolor/*/apps/*` but not if the icon defined in the main desktop file exist.

This plugs that hole.

See for example https://github.com/flathub/ml.mdwalters.Lemonade/pull/9#issuecomment-2069961956

The linter is saying:

```
    "errors": [
        "appstream-missing-icon-file",
        "appstream-missing-icon-key"
    ],
```

but the root cause is, the desktop file icon `ml.mdwalters.Lemonade` not existing in `files/share/icons/hicolor/*/apps/*`.

This is not caught by any other checks as the icon that is installed follows the exportable pattern: `ml.mdwalters.Lemonade.Devel.svg`


bbhtt:
Now https://github.com/flathub/ml.mdwalters.Lemonade/pull/9#issuecomment-2069961956 should look like, (which is better than before):

```json
{
    "errors": [
        "desktop-file-icon-not-installed",
        "appstream-missing-icon-file"
    ],
    "warnings": [
        "appstream-screenshot-missing-caption"
    ],
    "message": "Please consult the documentation at https://docs.flathub.org/docs/for-app-authors/linter"
}
```

PR #383, opened 2024-04-23
screenshot: Raise runtimeerror instead linter error on extraction fail

bbhtt:
We don't know what failed, so we cannot definitively say that appinfo dir is missing. RuntimeError is also consistent with usage everywhere else


bbhtt:
Never mind this doesn't quite work

PR #387, opened 2024-05-01
Add io.gitlab.liferooter.TextPieces to exceptions

liferooter:
I'm the developer of [Text Pieces](https://apps.gnome.org/Textpieces). I've just finished rewritting it. During the rewrite, it was moved from GitHub to GitLab with corresponding ID change. So now it's not `com.github.liferooter.textpieces`, but `io.gitlab.liferooter.TextPieces`. So I need to add it to the list of exceptions, because it still needs `--talk-name=org.freedesktop.Flatpak` permission.


bbhtt:
Do you have some docs/examples around what kind of commands it needs to run?

PR #391, opened 2024-05-04
Add io.github.elevenhsoft.WebApps to exceptions

elevenhsoft:
Hey,
I'm developing COSMIC Web Apps. It's allow to create "web applications" from your web browser of choice.
Unfortunately it's need `rw` permissions for creating desktop files and `ro` for reading what web browsers user have installed.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3227,5 +3227,8 @@
     },
     "net.rpcs3.RPCS3": {
         "module-rpcs3-source-git-branch": "Needed for custom updater to update once a day per upstream recommendation."
+    },
+    "io.github.elevenhsoft.WebApps": {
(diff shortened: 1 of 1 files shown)


bbhtt:
So looks like you are accessing the paths relative to $HOME and not `XDG_DATA_HOME` https://github.com/elevenhsoft/WebApps/blob/a8d5bc78d2a08c86d57760ac7ead550f6272e2c7/src/common.rs#L37-L42

You don't need `--filesystem=xdg-data/applications`, `--filesystem=~/.local/share/applications` is enough.

The difference is that `xdg-data/foo` binds mount the path and sets the environment variable `XDG_DATA_HOME`

elevenhsoft:
Oh, that's nice. So this exception is not needed? I will make an adjustment in that case.
What about the second one which is `xdg-data/flatpak:ro`?

elevenhsoft:
I just changed those paths so now it points to ~/.local/share instead of xdg-data. Sorry for confusions.
I did not need to use xdg-data since I'm using absolute paths in my code anyway. Ehh...

PR #396, opened 2024-05-16
Add exception for com.teamspeak.TeamSpeak

Zlopez:
This is to solve https://github.com/flathub/com.teamspeak.TeamSpeak/issues/57

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1554,7 +1554,8 @@
         "flathub-json-modified-publish-delay": "extra-data"
     },
     "com.teamspeak.TeamSpeak": {
-        "flathub-json-modified-publish-delay": "extra-data"
+        "flathub-json-modified-publish-delay": "extra-data",
+        "finish-args-unnecessary-xdg-config-access": "TeamSpeak 5 plugins installation directory."
     },
     "com.tencent.WeChat": {
         "finish-args-wildcard-kde-own-name": "Still uses legacy StatusNotifier implementation"


bbhtt:
I see `~/.config/TeamSpeak` was merged in https://github.com/flathub/com.teamspeak.TeamSpeak/pull/58

Do you still need this?

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -3229,4 +3229,7 @@
     "net.rpcs3.RPCS3": {
         "module-rpcs3-source-git-branch": "Needed for custom updater to update once a day per upstream recommendation."
     }
+    "com.teamspeak.TeamSpeak": {
```
bbhtt, line 3232: Missing `,` in the previous line

Zlopez:
> I see `~/.config/TeamSpeak` was merged in [flathub/com.teamspeak.TeamSpeak#58](https://github.com/flathub/com.teamspeak.TeamSpeak/pull/58)
> 
> Do you still need this?

Yes, the flathub builder doesn't allow me to publish the changes if there is no exception.

flatpak_builder_lint/staticfiles/exceptions.json
Zlopez: Fixed

bbhtt:
Ok then the issue you linked needs to be updated.

I see you are trying to solve https://github.com/flathub/com.teamspeak.TeamSpeak/issues/59

where the suggestion was to use `--persist`

Why are you switching to `xdg-config`?

Zlopez:
I was referencing to `~/.config/TeamSpeak` originally, so I though that using `xdg-config` will be more clean.
I'm still kind of confused by the permissions.

bbhtt:
If `--filesystem=~/.config/TeamSpeak` worked all you need is `--persist=.config/TeamSpeak`. This does not require any exceptions.

`xdg-config` is for when you want `~/.config/TeamSpeak` available under `$XDG_CONFIG_HOME`

Zlopez:
Hm, in this case let me try the persist first before merging this.

Zlopez:
It seems that changing to `--persist=.config/TeamSpeak` really makes the linter happy.

I'm closing this as it is no longer needed.

Zlopez:
I was too fast, the `--persist=.config/TeamSpeak` doesn't change the application to use `~/.config` as default`. So I tested it with `--filesystem=xdg-onfig/TeamSpeak` and it worked as it should.
That means that I really need the exception to be there.

hfiguiere:
What is it that you are trying to do?
`--persist` will make sure that whatever the app that do not follow the XDG spec write in that location WILL be written and persisted. This is the desired effect. This will keep it INSIDE the sandbox.

Zlopez:
The issue is that when installing add-ons or themes for TeamSpeak the installation scripts are using `.config/TeamSpeak/Default/extensions` which is not visible in flatpak TeamSpeak.

bbhtt:
> when installing add-ons or themes for TeamSpeak

Are these installed outside of the Flatpak? Manually or is there some kind of setup?

If it is manual it should be installed to `~/.var/app/com.teamspeak.TeamSpeak/config` instead.

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -3228,5 +3228,8 @@
     },
     "net.rpcs3.RPCS3": {
         "module-rpcs3-source-git-branch": "Needed for custom updater to update once a day per upstream recommendation."
+    },
+    "com.teamspeak.TeamSpeak": {
```
bbhtt, line 3232: There is an entry in https://github.com/flathub-infra/flatpak-builder-lint/blob/7ba6e745c3d14f3358a0a48d34b493d0db08a675/flatpak_builder_lint/staticfiles/exceptions.json#L1556, add this there.

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -3228,5 +3228,8 @@
...
+    	"finish-args-unnecessary-xdg-config-access": "TeamSpeak 5 plugins installation directory."
```
bbhtt, line 3233: There is an entry in https://github.com/flathub-infra/flatpak-builder-lint/blob/7ba6e745c3d14f3358a0a48d34b493d0db08a675/flatpak_builder_lint/staticfiles/exceptions.json#L1556

add the new exception there.

Zlopez:
> > when installing add-ons or themes for TeamSpeak
> 
> Are these installed outside of the Flatpak? Manually or is there some kind of setup?
> 
> If it is manual it should be installed to `~/.var/app/com.teamspeak.TeamSpeak/config` instead.

I found few that had their own installer on the TeamSpeak forum (for example https://community.teamspeak.com/t/skin-colorful-teamspeak-a-teamspeak-5-extension/24082)

So they are installed outside of flatpak by their own means. It will be easier for users to just read them from default location than doing a specific way for flatpak only.

flatpak_builder_lint/staticfiles/exceptions.json
Zlopez: I wasn't aware that there was already an entry for TeamSpeak, let me update it.

bbhtt:
Ok thanks I can now understand it. It'd be better if the scripts supported an extra install location for Flatpak-ed TeamSpeak.

Did `--filesystem=~/.config/TeamSpeak` solve the issue of reading addons? If so you can use that.

bbhtt:
> Did --filesystem=~/.config/TeamSpeak solve the issue of reading addons? If so you can use that.

If it doesn't and you need `~/.config/TeamSpeak` available under `$XDG_CONFIG_HOME` I'll merge this instead.

Zlopez:
What is the difference between `--filesystem=~/.config/TeamSpeak` and `--filesystem=xdg-config/TeamSpeak`. I tested it with the later, which worked.

bbhtt:
> What is the difference between `--filesystem=~/.config/TeamSpeak` and `--filesystem=xdg-config/TeamSpeak`. I tested it with the later, which worked.

`xdg-config/TeamSpeak` will bind mount `~/.config/TeamSpeak` to `$HOME/.var/app/com.teamspeak.TeamSpeak/config` and `$XDG_CONFIG_HOME=~/.var/app/com.teamspeak.TeamSpeak/config` inside the sandbox. If the application searches for the plugin path relative to `$XDG_CONFIG_HOME` then it needs that.

`--filesystem=~/.config/TeamSpeak` just gives a direct access. If the application is not XDG Dir aware and searches for the directory relative to `$HOME`, that one is enough.

Zlopez:
I need to test it with the `--filesystem=~/.config/TeamSpeak` then, so I can see if that is enough.

Zlopez:
It seems that the `--filesystem=~/.config/TeamSpeak` doesn't work.

bbhtt:
Ok thanks. Hard to verify what it is doing since it is closed source, but I assume this will work.

The exception will be up in a few hours. You can check if it is up: https://docs.flathub.org/docs/for-app-authors/linter#exceptions

Zlopez:
Thanks for help

PR #400, opened 2024-05-21
Add exceptions for dev.bsnes.bsnes

ghisvail:


bbhtt:
This should not be needed, if you are on Wayland the desktop theme is communicated to the sandbox via portals.

ghisvail:
I know. Sadly, the app uses a rather non-standard way of composing its GTK-based frontend.

They are not using portals, and they don't have the necessary skillset to make a proper refactoring to modern standards. It's basically a GTK-2 app with the bare minimal pieces of improvement to make it work with GTK-3.

bbhtt:
It doesn't need to use portals, the values are communicated regardless of them supporting portals or not. Just the system need to be running the portals.

bbhtt:
> It doesn't need to use portals, 

Actually I was thinking about a different theme option when I said this. The app indeed needs to read the values from settings portal or xsettings daemon.

I'll edit the reason and merge this.

PR #401, opened 2024-05-23
 exceptions: Add Subsurface to branch exception

Alexander-Wilms:
They use a custom updater via Github Workflow to update once per day as upstream creates builds on each commit.

> After a few years of doing release builds not on GitHub, we have come full circle, but in a different way.
> We now do "nightly" builds (they actually happen when our main branch changes) and post them in a separate nightly build repo

Cf. 

https://github.com/subsurface/subsurface/releases/tag/v4.9.4

and

https://subsurface-divelog.org/current-release/?lang=en

Also, unit tests are run after every Flathub build and prevent merging an update if they fail:

https://github.com/flathub/org.subsurface_divelog.Subsurface/blob/master/org.subsurface_divelog.Subsurface.yaml#L226

Regarding buildbot usage: The build only takes 10 min and I can just set the Actions workflow to be triggered once a week

image

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3182,6 +3182,9 @@
     "org.sparkleshare.SparkleShare": {
         "appstream-id-mismatch-flatpak-id": "appstream-cid predates this linter rule"
     },
+    "org.subsurface_divelog.Subsurface": {
+        "module-subsurface-source-git-branch": "Needed for custom updater to update once a day per upstream recommendation."
+    },
     "org.tabos.maxcontrol": {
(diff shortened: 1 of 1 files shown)


Alexander-Wilms:
Yes, weekly is fine.

bbhtt:
I guess this is not needed anymore as you found a workaround to using `branch` https://github.com/flathub/org.subsurface_divelog.Subsurface/blob/a3498c27a33c08cb92fa69ac7ae2c165289feb10/org.subsurface_divelog.Subsurface.yaml#L208-L213 but I'd still move it to once per week using a custom workflow.

We plan to deal with these applications doing daily builds later on - by moving them to weekly/bi-weekly etc..

PR #411, opened 2024-06-20
exceptions: allow automerge for com.fightcade.Fightcade

Pobega:
The Fightcade client needs to be in sync with the current server version to function. If an update causes breakages the end result is identical to having an outdated Flatpak (the client will not work in either case.)

Allow Flathubbot automerging to keep the client in sync even when a developer with merge access is unavailable.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1466,7 +1466,8 @@
         "flathub-json-modified-publish-delay": "extra-data"
     },
     "com.fightcade.Fightcade": {
-        "flathub-json-modified-publish-delay": "Client version must be in sync with server to function"
+        "flathub-json-modified-publish-delay": "Client version must be in sync with server to function",
+        "flathub-json-automerge-enabled": "Predates the linter rule, outdated clients cannot function"
     },
     "com.flashforge.FlashPrint": {
         "flathub-json-modified-publish-delay": "extra-data"


Pobega:
Note: automerge was removed from the repo in https://github.com/flathub/com.fightcade.Fightcade/commit/71fb933f68e3ff0c6f36656d10275c75eb46d76c to facilitate a Fightcade update (since it was blocking all pull requests due to the linter), but I would preferably like to restore that if an exception is granted.

bbhtt:
> The Fightcade client needs to be in sync with the current server version to function. I

Do you have a source for this? FAQ or bug reports etc. when it goes out of sync

bbhtt:
If you don't have it's fine, I can merge this on the basis of the description.

Pobega:
> Do you have a source for this? FAQ or bug reports etc. when it goes out of sync

It's the same as the other exception. Fightcade is a service that allows people to play with each other, once your client is out of sync (e.g. a new Fightcade releases, which is what we use the flathubbot PRs for) you will get an error like this (see attached image)

image

bbhtt:
It should be up in some time https://docs.flathub.org/docs/for-app-authors/linter#check-currently-published-exceptions

Pobega:
Thanks. Most bug reports for this client come through the Fightcade Discord, which tends to be a black hole for actually sourcing any issues. I'd prefer if people used the Github issue tracker, but it is what it is.

PR #417, opened 2024-06-26
exceptions: allow xdg-data access to net.codelogistics.webapps

eyekay-gh:
Web Apps installs websites as desktop applications, using xdg-desktop-portal's dynamic launcher portal for the same until now. However, I received multiple bug reports and complaints that the app was not working on various desktop environments and distros due to incomplete or outdated implementations of the portal and so the app now creates and removes desktop files in ~/.local/share/applications directly.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1217,6 +1217,9 @@
     "io.stoplight.studio": {
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
+    "net.codelogistics.webapps": {
+        "finish-args-arbitrary-xdg-data-access": "The app creates and removes desktop files in ~/.local/share/applications."
+    },
     "net.cozic.joplin_desktop": {
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },


bbhtt:
>  incomplete or outdated implementations of the portal

Was any bug opened to the portal folks? This really should be reported properly. 

Also curious to see what kind of bug reports from users you got.

eyekay-gh:
Hi, thank you for the prompt response!

[This](https://codeberg.org/eyekay/webapps/issues/11) is the bug report I got for KDE, which I reported to KDE [here.](https://bugs.kde.org/show_bug.cgi?id=488193)

[This](https://codeberg.org/eyekay/webapps/issues/15) is the bug report for Ubuntu 24.04, and [this](https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2069526) is the bug report I filed against this. The last time I checked, the fix for this was making its way through their packaging system.

[This](https://codeberg.org/eyekay/webapps/issues/20) seems to be related to the portal as well as I cannot reproduce it on my machine.

As you can tell it is not so much issues with the portal itself but with its various backends.

I was also told by a user I know IRL that they were unable to install a web app in Cinnamon but could do so in GNOME, but I was unable to investigate further.

bbhtt:
The KDE issue was fixed recently, https://github.com/flatpak/xdg-desktop-portal/pull/1376#issuecomment-2182121010

I'm granting the exception, but you should try to revert to the portal when possible.

eyekay-gh:
Hi, thank you so much! I'll switch back to the portal after the changes reach the major distros.

PR #420, opened 2024-07-01
Update exceptions.json

yairm210:
Added automerge exception for Unciv - averages to 2 releases a week

Unciv itself is build in a Github Actions pipeline - if you would rather I use the [Github action](https://github.com/flathub-infra/flatpak-github-actions) for flatpak releases I could do that, not sure if there are problems in "transferring" from the bot to the github action

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1621,6 +1621,9 @@
     "io.github.MakovWait.Godots": {
         "finish-args-flatpak-spawn-access": "flatpak-spawn is needed for blender and external code editors"
     },
+    "com.github.yairm210.unciv": {
+        "flathub-json-automerge-enabled": "Semiweekly release"
+    },
     "io.wavebox.Wavebox": {
         "flathub-json-modified-publish-delay": "extra-data"
     },


bbhtt:
What's blocking you from manually merging the two PRs?

yairm210:
I have 8 different distribution paths for my app, I don't have time or attention for manual work per release
If your new policy is that everything requires manual approval, that's also a valid answer, and I'll remove this app from Flathub entirely
(Not in like a bad threatening way, just practical considerations, don't mean to come off as rude, but I'd rather not have outdated versions lying around where people will use them and complain about ancient irrelevant problems)

bbhtt:
That was always the policy - the status checks, not pushing directly to main etc was for that. It just wasn't enforced and automerge allowed a way out. Automerge was introduced only for extra-data sources, but started being used as an all purpose automation.

Without a proper reason I can't really allow this exception, then pretty much everyone will say the same thing to get it.

yairm210:
OK, no problem, got it 👍🏿 
If I use the Github action, can that lead to a touchless deploy? Or would that require manual intervention as well?

bbhtt:
You can't really upload to Flathub with github-actions, direct-upload is only available to a few trusted publishers like Mozilla etc.

You can however host your own Flatpak repo, I have a bunch of examples here https://docs.flatpak.org/en/latest/hosting-a-repository.html#hosting-a-repository-on-gitlab-github-pages

yairm210:
Thanks for your help!

PR #422, opened 2024-07-03
exceptions: add dev.lizardbyte.dev.Sunshine

ReenigneArcher:
Sunshine requires the ability to start applications.

See: https://github.com/flathub/dev.lizardbyte.app.Sunshine/issues/36#issuecomment-2203990470

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1133,6 +1133,9 @@
     "dev.lapce.lapce": {
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
     },
+    "dev.lizardbyte.app.Sunshine": {
+        "finish-args-flatpak-spawn-access": "needed to start applications"
+    },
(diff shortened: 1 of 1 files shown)


bbhtt:
Can you please say what applications you need to starts and why?

ReenigneArcher:
Sunshine is a game streaming host for [Moonlight](https://moonlight-stream.org/).

The user can add any game or application they want as a launchable command. The user then gets a UI in the Moonlight client with posters to each application/game they added.

The most common applications to start are probably Steam, Lutris and other game launches, but many users also add individual games.

bbhtt:
Ok thanks, I'm assuming this is the docs for that https://docs.lizardbyte.dev/projects/sunshine/en/latest/about/setup.html#application-list

bbhtt:
it should be up in a few hours; you can check with https://docs.flathub.org/docs/for-app-authors/linter#check-currently-published-exceptions

ReenigneArcher:
Sorry for the delay. Yes those are the docs for adding applications.

Thank you!

PR #427, opened 2024-07-06
io.github.dweymouth.supersonic - Update exceptions.json

whorfin:
needs access to `xdg-config/pipewire:ro` for pipewire client config

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3426,7 +3426,8 @@
         "finish-args-contains-both-x11-and-wayland": "Predates the linter rule"
     },
     "io.github.dweymouth.supersonic": {
-        "finish-args-contains-both-x11-and-wayland": "Predates the linter rule"
+        "finish-args-contains-both-x11-and-wayland": "Predates the linter rule",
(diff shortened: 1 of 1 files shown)


bbhtt:
The maintainer is supposed to request exceptions cc @anarcat


Does `--filesystem=~/.config/pipewire:ro` work? Or you need it to be xdg-config specifically?

whorfin:
Yes, I was requesting access for the convenience of the maintainer
`~/.config/pipewire` works for my configuration, but the pipewire docs are explicit about the general case being `$XDG_CONFIG_HOME/pipewire/` for user configurations.

 https://docs.pipewire.org/page_man_pipewire-client_conf_5.html

bbhtt:
Then switch it to that and you'll need no exception. 

If someone keeps xdg-config on a different location, they are on their own to make changes to the sandbox.

whorfin:
Well OK then
I'm just trying to follow standards and make things easy
If you really object, I'll go back to the project and do the "mostly works" solution

bbhtt:
Also static permissions and sharing config from host, should be reduced as much as possible. There is very few people with custom pipewire and pulseaudio configs as most will just use distro defaults.

You can keep that as an override for yourself instead of introducing it for everyone.

PR #451, opened 2024-07-29
Automerge exception for Mattermost in beta branch

SemaiCZE:


bbhtt:
What's the reason you can't merge it manually?

Only verified apps were grandfathered, you cannot use "predates the linter rule here"/

SemaiCZE:
> What's the reason you can't merge it manually?

Of course I can, but I think that the `beta` branch is intended for testing, both unreleased (RC) versions and flatpak build updates. It's not guaranteed to be 100% stable as the `master` branch. I'm actively using the beta builds every day and it would be easier if it merges auomatically without my intervention. I'm still going to do all the testing in beta branch before master, so there is no point to forbid the automege option for that. I'll merge it and test it there anyways.

SemaiCZE:
> you cannot use "predates the linter rule" here

But it's true :slightly_smiling_face:. What would you like to read there?

bbhtt:
Flathub is not for nightly software, those nightly tags shouldn't be published daily to beta branch.

SemaiCZE:
> Flathub is not for nightly software, those nightly tags shouldn't be published daily to beta branch.

I'm not publishing dailies/nightlies, I'm publishing RC versions in `beta` branch and stable versions in `master` branch. You can see there are [not too many](https://github.com/flathub/com.mattermost.Desktop/blob/beta/com.mattermost.Desktop.appdata.xml) of them.

bbhtt:
If there's not too many, there shouldn't be any issue to merge them once or twice a week or so? Whether you test or not is up to you, there's no way to force that.

SemaiCZE:
It's sad that you're against automation and saving a little bit of human work, but it there is no way you would merge this we can live with that.

bbhtt:
> It's sad that you're against automation and saving a little bit of human work

The automation is working by sending PRs. It was never meant to be used to merge PRs without any human interaction like the blog post says for regular apps. The beta branch is also exposed to users.

If I grant this here without a proper reason, 10 more apps will come saying "this app got it for 'saving time', grant us too for the same reason". Then blocking it has no meaning. 

That's why I need a legitimate reason to give it, eg. your app breaks if not auto updated to the latest version like Discord, or upstream deletes old packages from the server etc.

PR #452, opened 2024-07-29
exceptions: ca.desrt.dconf-editor

yakushabb:
cid-missing-affiliation-gnome: Predates the linter rule

```
    "appstream": [
        "E: ca.desrt.dconf-editor.desktop:4: cid-missing-affiliation-gnome ca.desrt.dconf-editor.desktop"
    ],
```


bbhtt:
This was is also not a linter error, you have to remove the `project_group` tag from metainfo.

The `GNOME` value there can only be used by appids starting with `org.gnome.`. dconf-editor is in a weird position here even though it is a GNOME app the ID does not start with `org.gnome.`

PR #455, opened 2024-08-01
Port to ostree API and drop CLI usage

bbhtt:
TODO:

- [x] Restore checking if screenshot files are in the ref. Not sure if this is useful because it was just a warning since now.
- [x] Resolve these warnings:

```
** (process:116475): WARNING **: 19:59:01.778: expected enumeration type void, but got PyOSTreeRepoCheckoutMode instead

** (process:116475): WARNING **: 19:59:01.778: expected enumeration type void, but got PyOSTreeRepoCheckoutOverwriteMode instead
```

### DIFF
--- a/flatpak_builder_lint/ostree.py
+++ b/flatpak_builder_lint/ostree.py
@@ -1,69 +1,37 @@
-import errno
 import json
 import os
-import subprocess
-from typing import Optional, TypedDict
+from typing import List, Optional
 
-from .builddir import parse_metadata
+import gi
 
-
-class CliResult(TypedDict):
-    stdout: str
-    stderr: str
-    returncode: int
+gi.require_version("OSTree", "1.0")
+from gi.repository import Gio, GLib, OSTree  # noqa: E402
 
 
-def cli(repo: str, *args: str) -> CliResult:
-    if not os.path.exists(repo):
-        raise OSError(errno.ENOENT, f"No such ostree repo: {repo}")
+def open_ostree_repo(repo_path: str) -> OSTree.Repo:
+    repo = OSTree.Repo.new(Gio.File.new_for_path(repo_path))
 
-    ret = subprocess.run(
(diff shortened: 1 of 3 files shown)


bbhtt:
> Resolve these warnings:

Hey @TingPing any idea about these warnings? You can see the warnings by running `flatpak-builder-lint repo /path/to/repo` against this branch. The closes thing I got was this pygobject bug report https://gitlab.gnome.org/GNOME/pygobject/-/issues/608

TingPing:
Your usage looks correct to me, and ostree's GIR looks correct to me.

I believe it really is a pygobject bug. The recommended workaround of using `int()` works fine and is safe to do.

I would guess this is caused by `RepoCheckokutAtOptions` being a raw struct, this is uncommon and annoying for bindings, like `gjs` simply doesn't support it at all.

bbhtt:
Thanks!

bbhtt:
slightly changed `appstream-screenshots-files-not-found-in-ostree` and bumped it to a warning. It either is an error or doesn't stay at all, having it as a warning doesn't seem very useful.

Now I intersect between two sets: one set is filenames calculated from `screenshots/screenshot/image/text()` tag from catalogue and another set is filenames from extracting the screenshot ref for each arch.

If the intersection is empty, the error is produced.

flatpak_builder_lint/ostree.py
```diff
@@ -1,40 +1,32 @@
...
-    )
+    return repo
 
-    return {
-        "stdout": ret.stdout.decode("utf-8"),
-        "stderr": ret.stderr.decode("utf-8"),
-        "returncode": ret.returncode,
-    }
 
+def get_refs(repo_path: str, ref_prefix: str | None) -> set:
```
TingPing, line 20: Minor nit, you have functions returning a bare `set` and `dict`, instead of specifying the full type.

PR #466, opened 2024-08-14
Add exception for io.github.lime3ds.Lime3DS

OpenSauce04:
This application requires an exception for the `finish-args-unnecessary-xdg-data-applications-create-access` error as it has functionality that allows for creating game-specific shortcuts which appear in the host application menu, which involves writing to `$HOME/.local/share/applications`.


bbhtt:
The dynamic launcher portal exists for this functionality, it should use that.

PR #471, opened 2024-08-22
Add exception for Headlamp

joaquimrocha:
Headlamp needs to run flatpak-spawn for external k8s related tools.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3905,5 +3905,9 @@
     },
     "br.gov.fazenda.receita.irpf2024": {
         "external-gitmodule-url-found": "Predates the linter rule"
+    },
+    {
+    "io.kinvolk.Headlamp": {
+        "finish-args-flatpak-spawn-access": "Required to run external CLI tools related to Kubernetes. This is mainly to cover any tools the user has in their kubeconfig, or that Headlamp plugins need to run (e.g. minikube, az, etc.)."
     }
 }


bbhtt:
Can you not ship these tools directly inside the Flatpak? Openlens seems to does it https://github.com/flathub/dev.k8slens.OpenLens/blob/cbbb752bd660dda77df7e032ce4e357e478db923/dev.k8slens.OpenLens.yml#L48

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -3905,5 +3905,9 @@
     },
     "br.gov.fazenda.receita.irpf2024": {
         "external-gitmodule-url-found": "Predates the linter rule"
+    },
+    {
+    "io.kinvolk.Headlamp": {
+        "finish-args-flatpak-spawn-access": "Required to run external CLI tools related to Kubernetes. This is mainly to cover any tools the user has in their kubeconfig, or that Headlamp plugins need to run (e.g. minikube, az, etc.)."
     }
```
bbhtt, line 3912:
```suggestion
    },
    "io.kinvolk.Headlamp": {
        "finish-args-flatpak-spawn-access": "Required to run external CLI tools related to Kubernetes. This is mainly to cover any tools the user has in their kubeconfig, or that Headlamp plugins need to run (e.g. minikube, az, etc.)."
    }
```

joaquimrocha:
I could ship the tools that I am creating wrappers for (I know this approach does seem hacky), but there's always the case that someone may need add the full flatpak-spawn command to their kubeconfig in order for it to work with some tool. There are dozens of tools I can imagine people use, and a bunch that would be just increasing the bundle size as they would never use them. An alternative would be to build them as runtime extensions, but the UX for adding these was not the best, last time I checked.
Maintaining all of those though, seems overkill and a burden.
Even with those tools, I guess the app will need access to the user's home still. I know it's a different matter, but it's related.
Let me know what is the most appropriate, since I've not been on top of what the community has done in similar situations, and I know that it's hacky to have these wrappers anyway.

joaquimrocha:
@bbhtt Here is the related PR: https://github.com/flathub/io.kinvolk.Headlamp/pull/39

bbhtt:
flatpak-spawn assumes the tools are installed in host system, so it won't work if they aren't and you can't really check without failing first. I don't think you can assume that everyone installing the Flatpak will have aws or kubeconfig installed already.

The expectation is that, the Flatpak package is "complete" and doesn't depend on host stuff more than it needs to do.

bbhtt:
Size isn't a concern if the app needs it to function properly, you don't have to ship everything, just the ones that people have asked for. And most of them are just binaries you copy.

If the tools itself require running arbitrary host commands, then I think there's not an option here other than granting flatpak-spawn access.

joaquimrocha:
That's the thing, different users may need different tools and they will have their kubeconfig use them to e.g. authenticate against a public cloud.
The tools not existing is not a problem, since the command will fail similarly to what it would without the wrapper.
Having dozens of binaries installed for users who will never use them does seem like a waste of effort, since the users are more likely to have those external tools they need already before they have Headlamp.
Maybe one solution is to not include the flatpak talk name and recommend people to override that if they really need it some tool.

bbhtt:
Ok that's fine

> Maybe one solution is to not include the flatpak talk name and recommend people to override that if they really need it some tool.

If you can live with that it's fine by me.

joaquimrocha:
I am closing this for now since I think it should not be complicated to ask the users to grant the Flatpak talk name as extra. Thanks @bbhtt .

PR #478, opened 2024-09-21
Added exception for Audio plugins BaseExtension

hfiguiere:
-Added the exception to the validation list


bbhtt:
This won't work. `skip-appstream-check` in flathub.json has no effect anymore. It's left as a legacy parameter.

Checks are done or skipped automatically based on target ie. if it's a desktop/console app or a baseapp or an extension/runtime.

PR #479, opened 2024-09-21
Revert "Added exception for Audio plugins BaseExtension"

hfiguiere:
It doesn't work, target package removed what trigger the error as it is no longer relevant.

This reverts commit cf4f375e9c9332ccd8874ee64a2725c882eae884.


hfiguiere:
See #478

PR #485, opened 2024-09-28
Add source-git-branch exception-to pegasus frontend

j8r:
Likely what's the maintainer @mmatyas wish to (https://github.com/flathub/org.pegasus_frontend.Pegasus/issues/44).

Update: this is required for a weekly update x-checker cron (https://github.com/flathub/org.pegasus_frontend.Pegasus/pull/49)

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1453,7 +1453,8 @@
     },
     "org.pegasus_frontend.Pegasus": {
         "finish-args-flatpak-spawn-access": "the app predates this linter rule",
-        "flathub-json-automerge-enabled": "Predates the linter rule"
+        "flathub-json-automerge-enabled": "Predates the linter rule",
+        "module-pegasus-frontend-source-git-branch": "Needed for x-checker as upstream does not tag releases."
     },
     "org.sil.FieldWorks": {
         "finish-args-flatpak-spawn-access": "the app predates this linter rule"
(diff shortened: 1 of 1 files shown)


hfiguiere:
No. commit+branch mean the build WILL fail as soon as the branch HEAD changes. There is no going around. 

Flathub is NOT for nightly builds.

j8r:
There is commit and a branch, currently:
```yaml
 - type: git
   url: https://github.com/mmatyas/pegasus-frontend.git
   branch: master
   commit: cc2497abf7ebb12e0cbdb57a16addaf62d6f50bd
```
The goal is to update the commit to the latest one in master, which is the latest release.

bbhtt:
>  I guess an other release cycle will have to be found.

Yes please, one that doesn't involve updating daily to a new commit or tag or whatever.

If you want to do that, either update it with a custom workflow once a week like I mentioned in the original issue above or you are free to host it externally. 

Doing nightlies is not sustainable for the infra.

j8r:
Once a week sounds good, let's do like RPCS3. To have this https://github.com/flathub/org.pegasus_frontend.Pegasus/pull/49 work, an exception will still be needed though?
Or, we'll have to make a custom script for that purpose :thinking:

j8r:
If I understand correctly the issue, this is more about the frequency of builds than the nature of the release?
For example, a stable release per day is more problematic than a nightly build per month.

bbhtt:
Frequency is a problem for the infra as well as having software that is not "stable". End users expect software from Flathub to be stable, Nightly belongs in upstream CI.

mmatyas:
> @mmatyas do you confirm that the latest commit in master is considered a stable release and not a nightly one? Otherwise, I guess an other release cycle will have to be found.

The latest stable release is the last commit with an annotated Git tag (see https://github.com/mmatyas/pegasus-frontend/releases). I assume this would require a daily-running script, but if it's too much trouble, I'm fine with a weekly workflow too.

I also had good experience with externally hosting an auto-updated Flatpak release; we can set one up for nightlies if necessary.

bbhtt:
If you switch to those weekly_ tags, exception is not needed anymore.

j8r:
What do we do then @mmatyas, have a weekly cron taking the latest commit with the `continuous` tag?

mmatyas:
Yes, that sounds good to me.

PR #492, opened 2024-10-05
Checkout only directories needed by linter

bbhtt:
We have to ignore the "directories not existing in the refs"-type errors now as we don't know if they do but if the don't, the repo check aborts with Glib exception, so `True` is passed to extract_subpath.

This should be fine as we don't depend on directories itself for checks but specific files existing or not existing inside those directories.


bbhtt:
This is done because checking out `files/share` causes space to run out on the repo server for some apps.

PR #504, opened 2024-10-09
Add finish-args-flatpak-spawn-access to io.github.wivrn.wivrn

Meumeu:
WiVRn needs finish-args-flatpak-spawn-access to be able to automatically launch a game or a launcher of the user's choosing when a VR headset is connected.


bbhtt:
Is there any docs for this feature or can you point to the source code as a reference ?

PR #516, opened 2024-10-27
BoilR - Update exceptions.json

PhilipK:
These changes are needed to allow building BoilR, a tool that adds shortcuts to Steam.


bbhtt:
there is an entry for io.github.philipk.boilr on line 1275, can you merge it with that?

PhilipK:
> there is an entry for io.github.philipk.boilr on line 1275, can you merge it with that?

merged into a single entry now, thanks for spotting that.

PR #527, opened 2024-11-02
cosmicmoney exception

francesco-gaglione:


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -3833,6 +3833,9 @@
     "com.francescogaglione.chronos": {
         "finish-args-unnecessary-xdg-config-cosmic-ro-access": "Required to read the system theme and listen for config changes in the host"
     },
+    "com.francescogaglione.cosmicmoney": {
+        "finish-args-unnecessary-xdg-config-cosmic-rw-access": "Required to read the system theme and listen for config changes in the host"
```
bbhtt, line 3837: Why does it need to be rw? If you only need to read and listen it should be ro?

bbhtt: Same for the permission shouldn't it be `xdg-config/cosmic:ro`?

francesco-gaglione: Changed to ro, checked now I don't need it

PR #534, opened 2024-11-08
Update Tasks and Calculator permissions.

edfloreshz:
This PR adds write permission to `dev.edfloreshz.Tasks` and `dev.edfloreshz.Calculator` to be able to write app configuration files.


bbhtt:
why do app's config files go in `xdg-config/cosmic`? Shouldn't they be in app's own folder?

edfloreshz:
We initially granted read-only permissions to the `xdg-config/cosmic` directory to enable reading theme changes. However, we subsequently discovered that once any level of host access is provided, crates like `xdg` or `dirs` prioritize the host. 

This makes it challenging to write app configuration to the sandbox while simultaneously reading theme changes from the host, unless there's an alternative approach we've overlooked

edfloreshz:
I believe that will happen at some point, but in the meantime we still have to rely on this.

PR #539, opened 2024-11-14
Sioyek needs access to configuration files under xdg-config/sioyek/

nbenitez:
Sioyek uses his own directory xdg-config to look for the user theming configuration file "prefs_user.config". We need this to fix below bug: https://github.com/flathub/com.github.ahrm.sioyek/issues/24#issuecomment-2439694364

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3859,5 +3859,8 @@
     },
     "io.github.lawstorant.boxflat": {
         "finish-args-flatpak-spawn-access": "Required for automatic per-game preset application based on process name"
+    },
+    "com.github.ahrm.sioyek": {
+        "finish-args-unnecessary-xdg-config-sioyek-rw-access": "Required for app theming. Predates linter rule."
     }
 }


bbhtt:
Why is this needed? For the Flatpak the config file lives in `~/.var/app/$app-id/config`. I don't understand why you need host's config file.

nbenitez:
> Why is this needed? For the Flatpak the config file lives in `~/.var/app/$app-id/config`. I don't understand why you need host's config file.

Sioyek does not use xdg standard, it tries to read file `~/.config/sioyek/prefs_user.config` which fails inside flatpak, but if I run the flatpak with `flatpak run --filesystem=xdg-config/sioyek com.github.ahrm.sioyek your_file.pdf` then it can read it, hence why I opened this MR to incorportate that fix.

Do you have a better proposal for a fix?

nbenitez:
I don't think having to tell **all** users of an app that they have to run it with a specific flag for the app to work normal is a good standard.

Users (like you and I) want apps that work. I think that is the motivation for the exceptions file, to try make flatpak apps work as good as their original (non-flatpak) version.

If there's no a better suggestion, I look forward to this exception be merged.

bbhtt:
No, add that permission to the manifest instead of this.

bbhtt:
Or you can use `--persist` too.

nbenitez:
> No, add that permission to the manifest instead of this.

I already did that, and the build failed with:

```
    "errors": [
        "finish-args-unnecessary-xdg-config-sioyek-ro-access"
    ],
    "message": "Please consult the documentation at https://docs.flathub.org/docs/for-app-authors/linter"
```

And https://docs.flathub.org/docs/for-app-authors/linter tell me I needed to add an exception here to fix the build.

bbhtt:
`--filesystem=~/.config/sioyek` is not blocked in linter, not sure what you are doing.

nbenitez:
> `--filesystem=~/.config/sioyek` is not blocked in linter, not sure what you are doing.

Ok, I used `--filesystem=~/xdg-config/sioyek` locally and it worked so I didn't look further, then the build failed when try to upstream it.

So, I haven't tried `--filesystem=~/.config/sioyek` , which will probably work, will try that as it seems that is preferred than the exceptions file route.

hfiguiere:
`~/xdg-config` is not a thing. I mean it could exist, but it's not any kind of standard.

nbenitez:
> `~/xdg-config` is not a thing. I mean it could exist, but it's not any kind of standard.

Sorry, a typo, I meant `--filesystem=xdg-config/sioyek`

nbenitez:
It seems that both `--persist=.config` and `--persist=.config/sioyek` would work, so maybe https://docs.flathub.org/docs/for-app-authors/linter could also hint to use `--persist` before recommending to file exception file MR.

bbhtt:
You aren't supposed to use `--filesystem=xdg-config` if it doesn't follow `$XDG_CONFIG_HOME`, the docs don't know which one you need.

PR #547, opened 2024-12-09
Parse required-flatpak and enforce a minimum version for new metadata

bbhtt:
contingency plan if the fallback mechanism does not arrive in Flatpak 1.16.0


bbhtt:
Needs newer appstream, ubuntu-latest will switch to 24.04 next week.

bbhtt:
24.04 runners breaks docker arm64 builds, `cc` segfaults randomly. Likely OOM.

PR #569, opened 2025-01-27
exceptions: Allow qbittorrent to track libtorrent stable branch commits

Erick555:
See https://github.com/flathub/org.qbittorrent.qBittorrent/issues/175

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3875,5 +3875,8 @@
     },
     "app.freelens.Freelens": {
         "finish-args-flatpak-spawn-access": "required for host shell access via internal terminal"
+    },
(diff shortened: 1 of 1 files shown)


bbhtt:
Upstream doesn't seem low activity there were several commits done in the past week, with PRs open and failing builds.

Chocobo1:
>Upstream doesn't seem low activity there were several commits done in the past week, with PRs open and failing builds.

This is the branch we are talking about: https://github.com/arvidn/libtorrent/commits/RC_1_2
That branch has only seen a few commits (6 to be exact) in 2024. Libtorrent current main/default development branch is on `RC_2_0` or `master`.

Also it is worth to add that, for a long time, qBittorrent official releases are using libtorrent git commits and not stable releases. See: https://www.qbittorrent.org/news (expand the Library versions)

Therefore I support this PR and wish to add this exempt.

PR #580, opened 2025-02-20
Add `finish-args-contains-both-x11-and-wayland` exception for Discord

guihkx:
The client currently crashes if `--socket=wayland` and `--socket=fallback-x11` are used.

See https://github.com/flathub/com.discordapp.Discord/pull/492

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -2009,7 +2009,8 @@
     },
     "com.discordapp.Discord": {
         "appstream-id-mismatch-flatpak-id": "appstream-cid predates this linter rule",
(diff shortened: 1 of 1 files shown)


bbhtt:
Edited to include the PR link.

bbhtt:
This should be re-evaluated after a while. Most likely updating electron will fix it.

guihkx:
>This should be re-evaluated after a while.

Yeah, absolutely.

>Most likely updating electron will fix it.

If my interpretation of the gdb backtrace of the crash is correct (see the linked PR for that), it's a Discord module that causes the crash after querying for an X11 extension (`XInputExtension`) using an invalid X11 display handle.

If I had to wildly speculate why, that's because they probably need that extension for their global shortcuts feature.

PR #582, opened 2025-02-25
Update io.github.loot.loot exceptions

Ortham:
To cover the filesystem permissions needed for OpenMW support.


bbhtt:
can you provide some details? why do you need to access `/var/lib/flatpak/app/org.openmw.OpenMW:ro` and the `~/.var/app` paths?

Ortham:
> can you provide some details? why do you need to access `/var/lib/flatpak/app/org.openmw.OpenMW:ro` and the `~/.var/app` paths?

LOOT now supports reading and writing the load order of OpenMW plugins. For that, it needs to be able to read OpenMW's global config and builtin plugin, which are shipped inside the OpenMW flatpak, and needs to be able to read and write the plugin load order in the user-specific config, which is stored in `~/.var/app` when OpenMW is installed as a Flatpak.

PR #586, opened 2025-03-02
Add exception for page.codeberg.dnkl.foot

JustSoup312:
This PR adds an exception for `page.codeberg.dnkl.foot` to bypass `finish-args-flatpak-spawn-access` because it is a terminal and uses host-spawn to access the host's shell.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -10,6 +10,10 @@
     "org.flathub.exceptions_wildcard": {
         "*": "ignore all errors and warnings"
     },
+    "page.codeberg.dnkl.foot": {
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -10,6 +10,10 @@
     "org.flathub.exceptions_wildcard": {
         "*": "ignore all errors and warnings"
     },
+    "page.codeberg.dnkl.foot": {
+        "finish-args-flatpak-spawn-access": "required for host shell access via host-spawn"
```
bbhtt, line 14:
```suggestion
        "finish-args-flatpak-spawn-access": "required for host shell access via host-spawn",
```

JustSoup312: Ack, don't know how I missed that. Fixed now.

PR #591, opened 2025-03-06
Add exception for io.gitlab.payoliin.csnake

DodoLeDev:
[CSnake is an app in the process of being submitted to Flathub.](https://github.com/flathub/flathub/pull/6243)

However, because it's a TUI-based app, it is launched with `Terminal=true`, so graphical permissions are not needed


bbhtt:
Closing as submission PR is closed.

PR #593, opened 2025-03-07
Add external-gitmodule-url-found exception for OBS Studio

tytan652:


bbhtt:
This exception doesn't need to be submitted to us, we don't run this check on  our end for direct uploads.

PR #598, opened 2025-03-15
Add host tmp exception for Ungoogled Chromium

rany2:
In Chromium-based browsers, album art and video covers are stored in a temporary directory at `/tmp/.org.chromium.Chromium.<random_id>`. This path is then shared via the DBus property `org.mpris.MediaPlayer2.Player.Metadata`.

However, the `/tmp/` directory is not shared between the host and the Flatpak's namespace, preventing the host's notification area from displaying images in media controls.

Making the host's `/tmp/` directory writable resolves this issue.

https://github.com/flathub/com.brave.Browser/issues/750#issue-2921953257


bbhtt:
Does setting TMPDIR=/var/tmp work?

rany2:
Nope, Chromium ignores that.

PR #607, opened 2025-03-29
Add external-gitmodule-url-found exceptions

sonnyp:
I need to update my apps for GNOME 48, they all use the same git submodule 

I tried using a main module source instead of a git submodule unfortunately that fails with Builder and VSCode see https://github.com/bilelmoussaoui/flatpak-vscode/issues/251

I don't have time to fix this short term so I'd like to request these exceptions until I can figure it out.


bbhtt:
I don't see any git submodule in the Flathub repo? What submodule is this and where is it?

bbhtt:
Most likely you ran it inside your upstream repo checkout. Submodules are searched relative to git toplevel.

bbhtt:
Anyway I made the docs for this explicitly say manifest repo last week.

PR #641, opened 2025-05-11
Add exception for Reqable

JamesBrosy:


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1235,6 +1235,9 @@
         "finish-args-flatpak-spawn-access": "the app predates this linter rule",
         "finish-args-host-tmp-access": "Predates the linter rule"
     },
+    "com.reqable.Reqable": {
+        "flathub-json-automerge-enabled": "Predates the linter rule"
+    },
```
bbhtt, line 1240: a) it does not predate anything

b) the URL is versioned so it is not needed https://github.com/flathub/com.reqable.Reqable/blob/b81767368135818b96c23aaf5eab9740cf6d7b90/com.reqable.Reqable.yaml#L46

JamesBrosy: Is there any explanation for `automerge-flathubbot-prs`? In what cases is it necessary? Only the URL is versioned so it is not needed? I don't want to merge manually:(

bbhtt: It is granted if it is necessary on providing valid reasons, "I don't want to merge manually" is not a valid reason. Merging is a part of "maintenance"

JamesBrosy: I see. Thanks.

PR #646, opened 2025-05-15
Add finish-args-contains-both-x11-and-wayland exception for info.beyondallreason.bar

devurandom:
References: https://github.com/flathub/info.beyondallreason.bar/pull/51


bbhtt:
If it doesn't support ootb, then I don't see the point of having `--socket=wayland` anyone who need to manually set the env var can surely set the override too?

PR #648, opened 2025-05-18
Add "both x11 & wayland" exception for dev.vencord.Vesktop

Vendicated:
I listed some specific issues here https://github.com/flathub/dev.vencord.Vesktop/pull/55#issuecomment-2888697811 if it matters


bbhtt:
GitHub is having some issues, the exception will be up in a while.

PR #656, opened 2025-05-28
Add manifest-has-bundled-extension exception for com.fightcade.Fightcade

Pobega:
We have some work happening adding quality of life features for SteamOS that we want to implement using an optional com.fightcade.Fightcade.steamos extension.


Pobega:
There's a relevant PR up here: https://github.com/flathub/com.fightcade.Fightcade/pull/157

PR #661, opened 2025-06-10
Add io.github.swordpuffin.wardrobe

SwordPuffin:
The purpose of the permissions is to allow activating downloaded desktop themes inside of Wardrobe without the need of using an external utility.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1,6 +1,5 @@
 {
     "io.github.swordpuffin.wardrobe": {
-        "finish-args-flatpak-spawn-access": "needed to access host dconf configuration",
         "finish-args-dconf-talk-name": "needed to access host dconf configuration"
     },
     "io.github.TeamWheelWizard.WheelWizard": {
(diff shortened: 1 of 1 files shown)


SwordPuffin:
Flatpak-spawn is only needed in the initialization file (wardrobe.in).

It is not used anywhere else. My assumption is that it is necessary, I don't know if it works without flatpak-spawn.

https://github.com/SwordPuffin/Wardrobe/blob/1c2adb3af1f32a9748504135717b4ad4161fbdd9/src/wardrobe.in#L57C1-L61C14

SwordPuffin:
I need the schema path to write changes to the settings (icon-theme, gtk-theme, and cursor-theme) of the org.gnome.desktop.interface schema. 

The way wardrobe.in has been set up should filter out the irrelevant directories that are included in XDG_DATA_DIRS.

Also, I did some testing and I might be able to remove flatpak-spawn from permissions. It's currently being used to find  the XDG_DATA_DIRS path on the host machine rather than inside the sandbox, although I don't know if they would be different.

Sorry if I am not being clear, maybe the code might explain itself better:
https://github.com/SwordPuffin/Wardrobe/blob/1c2adb3af1f32a9748504135717b4ad4161fbdd9/src/wardrobe.in#L55C5-L83C77

bbhtt:
Yes if you can remove it it'd be better.

PR #665, opened 2025-06-16
Add io.github.Faugus.faugus-launcher

Faugus:
Without `socket=x11` Winetricks doesn't work on Wayland.
Without `socket=wayland` the Wine Wayland driver doesn't work.


Faugus:
@bbhtt What's wrong?

bbhtt:
it was merged nothing is wrong.

PR #672, opened 2025-06-22
Add org.firestormviewer.FirestormViewer

Quinn-Elara:
Added exception for dev-input, needed for gamepad support.


bbhtt:
input will not work on older Flatpak versions. Majority of the most used distros are on older Flatpak.

hfiguiere:
use `--device=all`. No exception. There will be a fallback (compatibility) mechanism one day in flatpak.

PR #677, opened 2025-06-30
Add com.jetbrains.Toolbox to exceptions.json

dreamyukii:


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1206,6 +1206,10 @@
         "external-gitmodule-url-found": "Predates the linter rule",
         "finish-args-host-filesystem-access": "Predates the linter rule"
     },
+    "com.jetbrains.Toolbox":{
+        "finish-args-home-filesystem-access":"Need full access to user homedir for function properly",
+        "finish-args-arbitrary-dbus-access":"Need dbus access for systray"
```
bbhtt, line 1211: what kind of systray requires full dbus access?

dreamyukii: unfortunately the systray is still broken even with another systray implementation like com.canonical.indicator.application and org.ayatana.indicator.application. so i guess having broken systray is the way to go.

PR #680, opened 2025-07-06
Add io.github.mpc_qt.mpc-qt

tbertels:
Access to home folder is needed for external subtitles and for downloaded mouse cursor themes from legacy ~/.icons path.
See https://github.com/mpc-qt/mpc-qt/issues/470 and https://github.com/mpc-qt/mpc-qt/issues/437

The app supports both XWayland and Wayland through a Tweaks setting, but both mode currently have their limitations, so it's better to let the user choose.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -6462,7 +6462,7 @@
         "finish-args-home-ro-filesystem-access": "Predates the linter rule"
     },
     "io.github.mpc_qt.mpc-qt": {
-        "finish-args-home-ro-filesystem-access": "Predates the linter rule"
+        "finish-args-home-ro-filesystem-access": "Predates the linter rule",
         "finish-args-contains-both-x11-and-wayland": "The app supports both X11 and Wayland through a Tweaks setting"
     },
     "io.github.punesemu.puNES": {


bbhtt:
What is the default on wayland? Just because it is configurable doesn't mean you should have both x11 and wayland sockets.

If the default is xwaylsnd then use x11 only.

bbhtt:
`~/.icons` is legacy it should not be used. You can put icon themes in ~/.local/share/icons

https://docs.flatpak.org/en/latest/desktop-integration.html#icons which is mounted in the sandbox by default.


Also is it possible to restrict subtitles etc. to xdg-videos and downloads only?

tbertels:
The default is x11, but users can use it in wayland mode on wayland.

Why prevent users of the flatpak version from doing the same?

tbertels:
> `~/.icons` is legacy it should not be used. You can put icon themes in ~/.local/share/icons
> 
> https://docs.flatpak.org/en/latest/desktop-integration.html#icons which is mounted in the sandbox by default.

It's not for me personally, it's for the users who don't choose where their cursor themes go, their DE does.
Why break that functionality for the flatpak version?

> Also is it possible to restrict subtitles etc. to xdg-videos and downloads only?

It's possible, but users then complain that the flatpak version is inferior to the normal version, and that we should warn them that not everything will work.

tbertels:
> Then they can add the permission ( `--socket=wayland` ) too.

That's not very user-friendly. Or is Flathub aiming only users with tech knowledge?

bbhtt:
You expecting non technical people changing the default is not user friendly.

tbertels:
> You expecting non technical people changing the default is not user friendly.

Well, it's just a checkbox in the options, versus an obscure command.

bbhtt:
> I can accept home because it needs to load playlists subtitles etc. But not x11 and wayland both.

and you already have an exception for `home:ro` 

https://github.com/tbertels/flatpak-builder-lint/blob/2302a43f8a563fd875e1fbca34e7e7609b819648/flatpak_builder_lint/staticfiles/exceptions.json#L6464-L6466

why is `home` r/w needed now?

bbhtt:
> Well, it's just a checkbox in the options, versus an obscure command.

Looks like it is setting `QT_QPA_PLATFORM`? You should let the toolkit decide instead of setting that. Qt6 defaults to wayland. There is no need to have an option for that.

bbhtt:
> It can't be :ro to access subtitles when a file is opened from the app.

Why? What write operations do you need to load subtitles?

tbertels:
> > Well, it's just a checkbox in the options, versus an obscure command.
> 
> Looks like it is setting `QT_QPA_PLATFORM`? You should let the toolkit decide instead of setting that. Qt6 defaults to wayland. There is no need to have an option for that.

In wayland mode, some features don't work, like keeping the window centered when it gets autoresized to fit the video in 100% size. This means that the player will be half hidden behind the visible screen.

tbertels:
> > It can't be :ro to access subtitles when a file is opened from the app.
> 
> Why? What write operations do you need to load subtitles?

It's mpv that does it. I've just noticed that :ro isn't enough, like for the limited permissions already granted in the manifest.

bbhtt:
> In wayland mode, some features don't work

Is that not more reason to default to xwayland and use `x11` socket only? 

If native wayland is broken for your app or doesn't offer the necessary features, why offer it as an option? Users will complain _more_ if they enable it and find things not working.

> It's mpv that does it. I've just noticed that :ro isn't enough

That's not sufficient explanation. Why does it need `write` it load subtitles?

bbhtt:
Even the MPV Flatpak doesn't have `home:rw`. It has 

r/w on `/tmp, xdg-pictures, xdg-videos`. and 

ro on `host`.

https://github.com/flathub/io.mpv.Mpv/blob/b96ee29bd0fb5c632c76e27ff853427364665bf6/io.mpv.Mpv.yml#L16-L20

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -6463,6 +6463,7 @@
     },
     "io.github.mpc_qt.mpc-qt": {
         "finish-args-home-ro-filesystem-access": "Predates the linter rule"
+        "finish-args-contains-both-x11-and-wayland": "The app supports both X11 and Wayland through a Tweaks setting"
```
bbhtt, line 6466:
```suggestion
        "finish-args-home-ro-filesystem-access": "Predates the linter rule",
        "finish-args-contains-both-x11-and-wayland": "The app supports both X11 and Wayland through a Tweaks setting"
```

tbertels:
Thanks!

PR #687, opened 2025-07-12
Add app.twintaillauncher.ttl exceptions

TukanDev:
File system home permission exception was mentioned here: https://github.com/flathub/flathub/pull/6704#discussion_r2193806961

As said multiple times I am confused why it can not reach perfectly online website address but I hope it is fine to exception that one too. You can check website is online and alive by visiting https://twintaillauncher.app

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1,7 +1,6 @@
 {
     "app.twintaillauncher.ttl": {
-        "finish-args-home-filesystem-access": "The application is a game launcher, it lets users choose where to store games, wine/proton etc... on their file system.",
-        "appid-url-not-reachable": "Website is alive and actually accessible??? Why linter can not reach it is beyond me, perhaps some reverse proxy issues???"
+        "finish-args-home-filesystem-access": "The application is a game launcher, it lets users choose where to store games, wine/proton etc... on their file system."
     },
     "com.ml4w.dotfilesinstaller": {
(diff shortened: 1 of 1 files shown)


TukanDev:
@bbhtt fixed website reach issue this should be mergeable now right?

TukanDev:
This exception is for /home access now only... It is literally needed for app to function regardless of weird session-bus case, also I did test abit ago file pickers seem to work fine but tauri-plugin-opener which has reveal item in directory crashes without it. I dont know why i annotated session bus perm was for file pickers now when i think about it.
But it is still an issue so I will see to investigate.

bbhtt:
> It is literally needed for app to function regardless

why?

TukanDev:
It lets users pick a directory?? To store game files and runners like wine etc... So they can pick their install locations, if there is a better way to do this then full home access do suggest it i am willing to fix it and not need an exception then, but i do not see any other logical way to grant permissions to files...

TukanDev:
Would it not need permission for file system access?? To even list you anything even thru file pickers??

bbhtt:
if the file pickers use portal you do not need any filesystem permission.

TukanDev:
See thats something I did not know atall... Will have to test without the permission to see if tauri plugin handles it using portals.
Sorry for inconvinience! First time releasing flatpak it just made logical sense i would need a permission even if its going thru a portal.

TukanDev:
Closing this as its proven not needed, thanks for pointing that out bbhtt

PR #689, opened 2025-07-13
eolruntime: Heuristically error on EOL runtimes

bbhtt:


bbhtt:
Fixes https://github.com/flathub/flathub/issues/2184

PR #691, opened 2025-07-14
Add `finish-args-contains-both-x11-and-wayland` exception for Avidemux

guihkx:
This seems to fix a start up crash affecting some Wayland users: https://github.com/flathub/org.avidemux.Avidemux/issues/29

We can't use `--socket=fallback-x11` at the moment, because that introduces a different start up crash (as detailed in the PR below)...

This is needed for: https://github.com/flathub/org.avidemux.Avidemux/pull/32

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -5077,7 +5077,8 @@
         "finish-args-host-filesystem-access": "Predates the linter rule"
     },
     "org.avidemux.Avidemux": {
-        "finish-args-host-filesystem-access": "Predates the linter rule"
+        "finish-args-host-filesystem-access": "Predates the linter rule",
+        "finish-args-contains-both-x11-and-wayland": "Fixes a start up crash for some Wayland users"
     },
     "org.beeref.BeeRef": {
         "finish-args-host-filesystem-access": "Predates the linter rule"


bbhtt:
If they don't want to support wayland why are you?

guihkx:
This is not really about supporting Wayland, though.

The main issue is that without `--socket=wayland` (i.e. using XWayland), the app crashes for some Wayland users (see the linked issue).

The idea here is just to prevent the crash for these users, and not "support Wayland".

guihkx:
Very well, then.

But just to avoid ambiguity on possible future requests, can you please clarify something for me?

Does upstream have a say on what we do with the package, even though they don't want to be involved and provide no support for it?

bbhtt:
Furthermore, I can't even reproduce the issue on Fedora 42 gnome and kde wayland VMs same as you. So we don't know what the actual issue is here, and at best this exception hides it.

guihkx:
I totally get what you're saying, but the status quo is not great either: The app is literally unusable for some people..

>Furthermore, I can't even reproduce the issue on Fedora 42 gnome and kde wayland VMs same as you. So we don't know what the actual issue is here, and at best this exception hides it.

I'll try to reproduce this on a real hardware and hopefully find the root cause.

bbhtt:
is this still needed? Seems like you found the cause?

guihkx:
Yup, sorry, I forgot about this.

PR #692, opened 2025-07-19
Add `finish-args-flatpak-system-folder-access` exception for SGDBoop

Eskander:
SGDBoop needs read-only access to `/var/lib/flatpak/app/com.valvesoftware.Steam` to check if Steam was installed using Flatpak.

https://github.com/flathub/com.steamgriddb.SGDBoop/pull/22

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3540,7 +3540,8 @@
     },
     "com.steamgriddb.SGDBoop": {
         "flathub-json-automerge-enabled": "Predates the linter rule",
-        "finish-args-flatpak-appdata-folder-access": "Predates the linter rule"
+        "finish-args-flatpak-appdata-folder-access": "Predates the linter rule",
+        "finish-args-flatpak-system-folder-access": "Read-only access to /var/lib/flatpak/app/com.valvesoftware.Steam to check whether Steam Flatpak is installed"
     },
     "com.steamgriddb.steam-rom-manager": {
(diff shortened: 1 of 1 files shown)


bbhtt:
Why do you need to check how steam was installed?

And checking for `/var/lib/flatpak...` to detect a Flatpaked steam is wrong

Eskander:
SGDBoop is used for customizing Steam libraries, and the library could either be located at `~/.steam/steam/` or `~/.var/app/com.valvesoftware.Steam/data/Steam/` depending on Steam installation method.

Earlier versions used to check `~/.var/app/com.valvesoftware.Steam/data/Steam/`, but this directory isn't cleaned by default when a user switches Steam installation method to a deb or rpm, so the check was switched to `/var/lib/flatpak...` to deal with this case.

bbhtt:
what you want is abandon flatpak and run as unsandboxed, none of those "detections" are correct.

bbhtt:
why does it matter or why are you detecting? have a priority-fallback system or ask the user about the installation method?

bbhtt:
> If this is deemed incorrect, then what is the upstream recommended method for detecting installed Flatpaks from inside a Flatpak?

flatpak info --show-location but obviously that won't work inside sandbox and same goes for detecting deb rpm or something other distro package manager.

Detecting the folder is not the same as detecting how steam is installed.

What you are essentially trying to do doesn't work in a sandbox.

bbhtt:
> Falling back to either location without detecting the current Steam installation method risks changing the wrong, left over library.

That is why you should ask the user where their steam library is.

PR #693, opened 2025-07-24
Add `finish-args-host-filesystem-access` exception for Clapgrep

luleyleo:
With read-only access, the GTK FileDialog still returns "Document API" style paths rather than the real path. This means Clapgrep can't display the actual path of the search directory or searched files. I don't understand why it only works with write access, but I would guess this is some kind of issue with Flatpak.

https://github.com/flathub/de.leopoldluley.Clapgrep/pull/15


bbhtt:
There is a portal API to get the real paths called GetHostPaths. Have you considered supporting that?

bbhtt:
It is yes, how important is that feature? It's rather advise you to not make the default and instead ask the user to enable the permission.

It essentially makes the app fully unsandboxed .

PR #697, opened 2025-08-09
Update exceptions for WiVRn

Meumeu:
This allows WiVRn to get the list of applications from .desktop files and Steam (https://github.com/WiVRn/WiVRn/blob/master/common/application.cpp)

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -4366,7 +4366,9 @@
     "io.github.wivrn.wivrn": {
         "finish-args-unnecessary-xdg-config-openvr-create-access": "Required to set the current OpenVR runtime",
         "finish-args-unnecessary-xdg-config-openxr-create-access": "Required to set the current OpenXR runtime",
-        "finish-args-flatpak-spawn-access": "Required to launch games on the host"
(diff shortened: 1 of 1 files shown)


bbhtt:
What permission are you adding and where is that PR? If you are adding `xdg-data/Steam:ro` then the exception should be for `finish-args-unnecessary-xdg-data-Steam-ro-access`.

xytovl:
xdg-data is used to read .desktop files in `~/.local/share/applications` as well as games from `/.local/share/Steam`

For the `--filesystem=host-os:ro` part, as far as I know it is the most granular permission we can use to read `/usr/share/applications` as `/run/host/usr/share/applications` and it is not possible to know the real values from `$XDG_DATA_DIRS` on the host. The `/run/host/share` directory only contains icons.

bbhtt:
Then two folders in xdg-data only still doesn't require the whole xdg-data. So the exception is for the wrong string.

xytovl:
What are the required changes?

In the flatpak manifest replace
```
--filesystem=xdg-data:ro
```
with
```
--filesystem=xdg-data/applications:ro
--filesystem=xdg-data/Steam:ro
```
and what in the exceptions?

bbhtt:
Yes

I posted the string a comment above or you can run the linter and see.

PR #698, opened 2025-08-10
Add exception for io.github.sockeye_d.godl

sockeye-d:
My Godot project and version manager godl needs host filesystem access since it creates, reads, and modifies Godot projects from the filesystem.

Is it okay to do this before creating a submission to Flathub?


bbhtt:
where is the submission? and why did you submit the exception without the app?

sockeye-d:
I thought the exception needed to be submitted before the app since the exception would block the linter from reporting the erroneous error? This is my first time submitting a flatpak so I'm not entirely sure of the process

bbhtt:
Open the PR with the app submission first, we don't even know whether we will accept the app.

sockeye-d:
Here's the PR: flathub/flathub#6799. The CI builds fail because it doesn't have the exception right now

PR #701, opened 2025-08-22
Add release-not-in-order exception for org.garudalinux.firedragon

stefanwimmer128:
The FireDragon project uses [SemVer](https://semver.org/) versioning and the releases in the MetaInfo are ordered accordingly (v12.0.0-beta.21 << v12.0.0-beta.21). But AppStream release order differs and thinks that v12.0.0 << v12.0.0-beta.21 (https://github.com/flathub-infra/vorarbeiter/actions/runs/17147357555/job/48646095336) which is wrong in SemVer. So I would like to request an exception for this linting rule to keep the currently sensible release order in MetaInfo.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -6492,6 +6492,9 @@
     "org.freedesktop.GstDebugViewer": {
         "finish-args-home-ro-filesystem-access": "Predates the linter rule"
     },
+    "org.garudalinux.firedragon": {
+        "releases-not-in-order": "This project uses SemVer which differs from AppStream release order"
+    },
     "org.gnome.Evince": {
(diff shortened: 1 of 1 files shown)


bbhtt:
This is not an error in the linter, it comes from appstreamcli validate. You should dispute it there https://github.com/ximion/appstream/

stefanwimmer128:
I reported the issue upstream. But I would still like to request this exception for now (temporarily until upstream issue is resolved). Since FireDragon v11 is based on older Firefox ESR128 which will be EOL soon (September 16th). And I won't be able to release v12 which is based on the latest Firefox 142 until this is resolved.

stefanwimmer128:
Ah, ok, I understand. Thanks for the clarification.

Upstream has just responded and provided a workaround, by replacing the `-` with `~` in the versions to be ordered correctly.

PR #708, opened 2025-08-30
io.github.kolunmi.Bazaar: Update exceptions.json

linux-universe:
https://github.com/flathub/io.github.kolunmi.Bazaar/pull/4


tulilirockz:
This is indeed functional, tested on my machine, would be good to get merged!

tulilirockz:
Actually seems that we need to do some more work on this

bbhtt:
let's close this, feel free to reopen when the app supports it.

PR #719, opened 2025-09-13
Add exceptions for io.github.shonubot.Spruce

shonubot:
Spruce is a GTK4/libadwaita system cleaner that helps users reclaim disk space by scanning and removing application cache files.  
To support its core functionality.

**Requested exceptions:**
- finish-args-arbitrary-xdg-cache-rw-access  
  → Needed to scan and clean user application caches in $HOME/.cache with explicit preview and confirmation.

- finish-args-host-ro-filesystem-access  
  → Allows read-only inspection of host mount info to calculate accurate disk usage before/after cleaning.  
    Cleaning itself remains limited to user-writable locations.

- finish-args-flatpak-spawn-access  
  → Required to invoke host tools (du/find/stat, etc.) for size checks and safe deletions, restricted to user-selected targets.

These permissions are essential for Spruce’s purpose as a system optimiser.
The app is yet to be submitted to Flathub [here](https://github.com/flathub/flathub/pull/6922)

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -6938,5 +6938,10 @@
     },
     "life.bolls.bolls": {
         "finish-args-full-home-config-access": "Predates the linter rule"
+     },
+    "io.github.shonubot.Spruce": {
+        "finish-args-arbitrary-xdg-cache-rw-access": "Required to scan and clean app caches in $HOME/.cache, with user preview and confirmation.",
+        "finish-args-host-ro-filesystem-access": "Needed for accurate disk usage calculations using host mount info; cleaning is limited to user areas.",
+        "finish-args-flatpak-spawn-access": "Allows calling host tools for size checks and safe deletions, restricted to user-selected targets."
     }
 }


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -6948,7 +6948,12 @@
     "life.bolls.bolls": {
         "finish-args-full-home-config-access": "Predates the linter rule"
     },
-    "it.mq1.TinyWiiBackupManager": {
+    "io.github.shonubot.Spruce": {
+      "finish-args-arbitrary-xdg-cache-rw-access": "Required to scan and clean app caches in $HOME/.cache, with user preview and confirmation.",
+      "finish-args-host-ro-filesystem-access": "Needed for accurate disk usage calculations using host mount info; cleaning is limited to user areas.",
+      "finish-args-flatpak-spawn-access": "Allows calling host tools for size checks and safe deletions, restricted to user-selected targets."
```
bbhtt, line 6954: what host tools are used for the cleanup now?

bbhtt: why `rm`? Python has shutil

shonubot: I switched host deletions to Python (`shutil`). With `--filesystem=xdg-cache:rw`, `~/.cache` is handled in-sandbox. For `~/.var/app/*/cache`, I can either (a) grant write access (e.g., `--filesystem=~/.var/app:rw`) and stay Python-only, or (b) keep a narrow host fallback (`rm`) just for that path. Happy to proceed with whichever you prefer.

shonubot: I updated the filesystem permission and `app.py` to only require the `flatpak-spawn-access` permission and now all cache directories can be accessed without any filesystem permissions.

PR #723, opened 2025-09-17
Allow full filesystem access for page.codeberg.cozyowl.OpaqueFiles

cozyOw1:
OpaqueFiles is a file encryption program primarily for use in combination with a cloud provider.

It needs to encrypt/decrypt files all over the filesystem, otherwise users will get scared when they are unable to decrypt their files due to OpaqueFiles being unable to find those files inside it's sandbox.

Simple read/write access (for a single file) trough a file chooser will not be enough, because to encrypt test.txt, test.txt will need to be replaced by it's encrypted version test.txt.of.

**Requested Exception**
- finish-args-host-filesystem-access - Allows for replacing encrypted/unencrypted files by their processed counterpart. Full file system access to avoid panicking users when they try to decrypt outside of HOME.

The program was not submitted to Flathub yet.


hfiguiere:
also don't merge branches. This should be a single commit.

PR #737, opened 2025-10-16
add finish-args-contains-both-x11-and-wayland for teamspeak

SenseiHiraku:
See https://github.com/flathub/com.teamspeak.TeamSpeak/pull/83#issuecomment-3410490310
Any other socket combination is causing a crash when using the newest version

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -6191,5 +6191,8 @@
     "io.github.jmberesford.Retrom": {
         "finish-args-flatpak-spawn-access": "the app is a launcher for other arbitrary applications configured by the user and requires the ability to spawn external processes on the host",
         "finish-args-home-filesystem-access": "many users place/install games and portable emulators/applications that Retrom must manage in arbitrary locations in the home directory"
+    },
+    "com.teamspeak.Teakspeak": {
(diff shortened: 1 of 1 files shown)


bbhtt:
why is it trying to use wayland when it is not available?

SenseiHiraku:
> why is it trying to use wayland when it is not available?

This is in the changelog @bbhtt 

 > "Added support for Pipewire Capturing under wayland

Wayland is needed in order to stream wayland applications, for which they've added support in the new version.
Teamspeak 6 is still in beta, so things might change in future, and then I'll let you know when the `finish-args-contains-both-x11-and-wayland` exception is not needed any more and things get more stable.

bbhtt:
Yes but that should not be used when it has no access to the wayland socket.

You cannot use wayland specific features when the app itself has no access to the wayland socket.

SenseiHiraku:
Even with `--socket=fallback-x11`, the application still fails to start. They've added the following message in the [new update today](https://community.teamspeak.com/t/teamspeak-6-0-0-beta3-3-minor-update/62656):
Bildschirmfoto_20251210_032342 @bbhtt 

Even with --socket=fallback-x11, the application still fails to start.
As mentioned before, using only socket=wayland allows the app to run, but it crashes as soon as the “Hotkey” settings are opened, since it apparently still relies on X11. Only the combination of `socket=wayland` and `socket=x11` provides stable operation.

Users of the AUR package or the direct download (running through XWayland) are not experiencing these issues, so it would be a shame if the Flatpak version were artificially limited. For this reason, I’m proposing an exception here so we can finally update the application.

PR #744, opened 2025-11-10
Add exception for io.trezor.suite

Ondra-Zik:
needed for https://github.com/flathub/io.trezor.suite/issues/20
The app expects Desktop Enviroment's or Blueman BT UI to pair new BT devices


Ondra-Zik-SL:
> What exactly is "host Bluetooth pairing UI"? What are you doing?

The app relies on desktop environment's built-in Bluetooth pairing to show and accept/decline the pairing code.

For example, on GNOME it opens `gnome-control-center`, and on KDE it opens `systemsettings5`. Without this, you cannot pair a new Bluetooth device from the app.

PR #756, opened 2025-11-18
manifest: Check for debug cleanup in subdirs and in modules

hfiguiere:
There are many packages that do it...

Expand the linter rule to check the modules cleanups
And to check for subdirs too.

### DIFF
--- a/flatpak_builder_lint/checks/modules.py
+++ b/flatpak_builder_lint/checks/modules.py
@@ -86,7 +86,7 @@ class ModuleCheck(Check):
         cleanup = module.get("cleanup")
         if cleanup:
             for c in cleanup:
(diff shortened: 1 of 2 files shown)


flatpak_builder_lint/checks/modules.py
```diff
@@ -83,6 +83,13 @@ def check_module(self, module: dict[str, Any]) -> None:
             for nested_module in nested_modules:
                 self.check_module(nested_module)
 
+        cleanup = module.get("cleanup")
+        if cleanup:
+            for c in cleanup:
+                if c == "/lib/debug" or c.startswith("/lib/debug"):
```
bbhtt:
```suggestion
                if c == "/lib/debug" or c.startswith("/lib/debug/"):
```

PR #757, opened 2025-11-18
Update exceptions.json for io.github.brunoherbelin.Vimix

brunoherbelin:
Application needs access to session-bus to inhibit the desktop screensaver via the org.freedesktop.ScreenSaver D-Bus API (live performance)


razzeee:
Why can't you use https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Inhibit.html ?

brunoherbelin:
Okay, I changed my code. thanks for pointing out to the right direction.

PR #759, opened 2025-11-20
Update IDEA, WebStorm and Android Studio

xanscale:


bbhtt:
Put an actual reason instead of "predates the linter rule".

xanscale:
done, but i continue to not really understand ssh-auth and gpg-agent permission.
i can fully use git/ssh without them

PR #765, opened 2025-11-21
Add exception for com.github.wwmm.easyeffects

vchernin:
Allow x11 and wayland support as x11 is needed for audio plugin windows, even though the application itself runs under Wayland.


bbhtt:
So the port introduced new bugs like needing arbitrary wrong combination of permissions. This will not be given.

PR #766, opened 2025-11-21
Add exception for moe.launcher.an-anime-game-launcher

NelloKudo:
Needed for Wine to be able to run on XWayland by default as intended by the Wine project, given that winewayland is still unstable.

See Bottles' example: https://github.com/flathub-infra/flatpak-builder-lint/blob/03f8008d97f291683830e2bd28dc1c886b274c95/flatpak_builder_lint/staticfiles/exceptions.json#L3451

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1,4 +1,7 @@
 {
+    "moe.launcher.an-anime-game-launcher": {
+        "finish-args-contains-both-x11-and-wayland": "Predates the linter rule"
+    },
     "app.twintaillauncher.ttl": {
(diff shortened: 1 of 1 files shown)


bbhtt:
Then you should use `--socket=x11` only

NelloKudo:
That would prevent Wine’s wayland driver to work at all though, I’d like to leave the final choice to the user in the end. There are various setups where one or the other would bring benefits, hence why I think both should work.

See https://github.com/flathub-infra/flatpak-builder-lint/pull/665, this falls under the exact same use case.

RushingAlien:
If i may, for game launcher apps this usecase is quite common, for example bottles is a GTK4 app, which preferably runs in Wayland if available, but Wine only officially supports x11 with an experimental wayland driver. This app is a game launcher so it would fit the exception as most game launchers are in the exception list already.

PR #768, opened 2025-11-23
Add D-Bus access for keyforge_master app

5t4r-10rd:
This app is a macropad configurator, it uses node serialport. It needs to have access to session bus to run


5t4r-10rd:
@bbhtt what's the reason for blocking this?

bbhtt:
you have to find the specific talk or own names needed. session-bus is not going to be allowed https://docs.flatpak.org/en/latest/sandbox-permissions.html#d-bus-access

PR #778, opened 2025-12-01
Add exception for chat.simplex.simplex

shumvgolove:
Due to framework limitations of Java/Compose Multiplatform, SimpleX Chat cannot currently utilize the XDG Desktop Portal API.


shumvgolove:
Thanks!

epoberezkin:
@bbhtt thank you. I think that early users were able to change this configuration themselves, and now we have a rather large number of "not working" complaints, as it gets adopted wider...

Thank you.

PR #780, opened 2025-12-02
Add exception for io.github.d0ksan8.zen-app-manager

d0ksan8:
Required for flathub/flathub#7229. This app is a Startup Manager and needs direct access to autostart directories.


bbhtt:
As I said there what is `xdg-data/autostart` for? xdg-config/autostart is the standard location

PR #781, opened 2025-12-03
Add exceptions for com.bilingify.readest

chrox:
Required for https://github.com/flathub/com.bilingify.readest/pull/4


bbhtt:
I see, you only need the own name. own-name implies talk access.

PR #785, opened 2025-12-08
Add io.sourceforge.xtrkcad_fork.xtrkcad exception

strobelight:
The xtrkcad app needs home directory access for reading/writing user track plans that could be stored anywhere on home directory.

Builds currently fail for pending PR: [pull#7255](https://github.com/flathub/flathub/pull/7255) without this exception.


bbhtt:
Added

PR #791, opened 2025-12-14
Add exceptions for io.github.nikelaz.CTLDash

nikelaz:
This is related to my [CTLDash app submission to Flathub](https://github.com/flathub/flathub/pull/7316).

The exceptions are described and are necessary as this is a systemd management application and does need some permissions to manage the user and system services.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1,6 +1,6 @@
 {
     "io.github.nikelaz.CTLDash": {
-        "finish-args-systemd1-talk-name": "Needed to list, and control (start, top, restart, enable, disable) systemd user services.",
+        "finish-args-systemd1-talk-name": "Needed to list, and control (start, stop, restart, enable, disable) systemd user services.",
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1,4 +1,10 @@
 {
+    "io.github.nikelaz.CTLDash": {
+        "finish-args-systemd1-talk-name": "Needed to list, monitor, and control user systemd services (start, stop, restart operations on user services).",
+        "finish-args-systemd1-system-talk-name": "Needed to communicate with systemd system daemon via D-Bus to list, monitor, and control system-wide systemd services.",
+        "finish-args-arbitrary-dbus-access": "Need both system and session D-Bus access to manage both system-level and user-level systemd services through the org.freedesktop.systemd1 interface.",
```
hfiguiere: this is not what you think it is. The very short version unless you have an actual justification to have blank dbus access (like this is a dbus introspector tool) it will not be granted.

nikelaz:
Absolutely, I apologize.

I removed the unnecessary permissions and tested the ones left to make sure they are exactly what I expect them to be.

PR #798, opened 2025-12-16
Add exception for org.kde.kjournaldbrowser

cordlandwehr:
kjournaldbrowser is a log viewer primarily for the system's journal log at /var/log/journal , thus access to that location is needed.

The manifest file is available here: https://invent.kde.org/system/kjournald/-/blob/master/.flatpak-manifest.yaml


bbhtt:
use https://github.com/flathub-infra/flatpak-builder-lint?tab=readme-ov-file#local-exceptions for your CI. these exceptions are for apps on flathub or being submitted.

PR #813, opened 2025-12-25
Clean some entries related to end-of-life applications

yakushabb:
It would be useful to have another way that cleans up the relevant entries for applications marked as EOL.


bbhtt:
There is a script to do this. I run it from time to time.

PR #815, opened 2025-12-27
Add new exception for goverlay introduced with 1.6.7 update

Twig6943:
I'll update the manifest to 1.6.7 shortly

Edit: I've updated the manifest


bbhtt:
Why is apps' mangohud config inside ~/.var/app? Shouldn't everyone write on host?

benjamimgois:
Hey @bbhtt , the traditional mangohud config file `~/.config/MangoHud/MangoHud.conf` only works for the native steam client. To use flatpak steam mangohud this is the correct file location. This is a old feature request that is being adressed by this.

https://github.com/benjamimgois/goverlay/issues/30

PR #818, opened 2025-12-28
Add exception for io.github.totoshko88.RustConn

totoshko88:
## Summary
RustConn is an SSH/RDP/VNC/SPICE connection manager for Linux (GTK4/libadwaita).

### `finish-args-ssh-filesystem-access`
RustConn needs read-only access to `~/.ssh` to load user's existing SSH keys and config files for SSH connections. This is essential for the app's core functionality as an SSH client.

### `finish-args-has-socket-ssh-auth`
Required for SSH agent forwarding to enable key-based authentication for SSH connections without re-entering passphrases.

## Links
- GitHub: https://github.com/totoshko88/RustConn
- Flathub PR: https://github.com/flathub/flathub/pull/7401

### DIFF
--- a/flatpak_builder_lint/appstream.py
+++ b/flatpak_builder_lint/appstream.py
@@ -3,9 +3,13 @@ import re
 import subprocess
 from typing import TypedDict, cast
 
+import gi
 from lxml import etree
 
 from . import config
+
+gi.require_version("AppStream", "1.0")
+from gi.repository import AppStream  # noqa: E402
 
 
 class SubprocessResult(TypedDict):
@@ -80,7 +84,7 @@ def is_developer_name_present(path: str) -> bool:
 
 
 def is_project_license_present(path: str) -> bool:
-    return is_present(path, "//project_license/text()")
+    return get_project_license(path) is not None
 
 
 def has_icon_key(path: str) -> bool:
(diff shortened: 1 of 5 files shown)


totoshko88:
> Are you generating the PR with LLMs?

First time it happened like this, the LLM sorted all items alphabetically. That exception was closed above. Now I have manually added only the block for the exception required for my app

bbhtt:
> Now I have manually added only the block for the exception required for my app

So the PR description was LLM generated?

hfiguiere:
> and it automatically added details about similar applications based on its analysis

It's not similar applications that are requesting the exception.

So I guess we can close this then.

totoshko88:
> > and it automatically added details about similar applications based on its analysis
> 
> It's not similar applications that are requesting the exception.
> 
> So I guess we can close this then.

I have removed the comments about similar solutoions. 
The application is a connection manager that effectively acts as a frontend for embedded functionality based on open-source Rust implementations (ironrdp, vnc-rs, etc.) or external clients (freerdp, tigervnc, etc.). However, the core functionality is built around VTE and the terminal for use with SSH, cloud CLIs, etc., where SSH serves as the primary authentication method. Consequently, two exceptions are required to access the default paths for SSH and the SSH agent. Without these exceptions, the application's core functionality will not work properly.

PR #819, opened 2025-12-28
exceptions: allow finish-args-own-name-org.opendeck.SingleInstance for me.amankhanna.opendeck

nekename:
https://v2.tauri.app/plugin/single-instance/#usage-in-snap-and-flatpak

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1792,7 +1792,8 @@
         "finish-args-unnecessary-xdg-data-gnome-background-properties-create-access": "https://github.com/dusansimic/dynamic-wallpaper/issues/5"
     },
     "me.amankhanna.opendeck": {
-        "finish-args-flatpak-spawn-access": "host shell access required for running user-provided commands and also to start plugins using Wine and Node"
+        "finish-args-flatpak-spawn-access": "host shell access required for running user-provided commands and also to start plugins using Wine and Node",
+        "finish-args-own-name-org.opendeck.SingleInstance": "the Tauri single instance plugin publishes this service to advertise the current instance to new instances"
     },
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -1792,7 +1792,8 @@
         "finish-args-unnecessary-xdg-data-gnome-background-properties-create-access": "https://github.com/dusansimic/dynamic-wallpaper/issues/5"
     },
     "me.amankhanna.opendeck": {
-        "finish-args-flatpak-spawn-access": "host shell access required for running user-provided commands and also to start plugins using Wine and Node"
+        "finish-args-flatpak-spawn-access": "host shell access required for running user-provided commands and also to start plugins using Wine and Node",
+        "finish-args-own-name-org.opendeck.SingleInstance": "the Tauri single instance plugin publishes this service to advertise the current instance to new instances"
```
hfiguiere: Why doesn't it use something under `me.amankhanna.opendeck`?

hfiguiere: That doesn't make it right. It's completely wrong and also violate d-bus convention and such.

Tauri should be fixed.

bbhtt: Open an issue to Tauri.

nekename: In fact, `com.bilingify.readest` and `dev.mrquantumoff.mcmodpackmanager` already have exceptions added for this, so I see no reason it can't be approved for my app.

nekename: Besides, it can't use the Tauri app ID, because that isn't always a reverse-DNS ID - mine is simply `opendeck`.

PR #821, opened 2025-12-29
Add lint exception for antimicrox

pktiuk:
Lint exception for AntiMicroX linked with: https://github.com/AntiMicroX/antimicrox/issues/1201


bbhtt:
You arern't going to get this exception. If it doesn't run with fallback-x11+wayland then use `--socket=x11` only.

Any other combination is plain wrong and a bug in the app which needs to be fixed.

Erick555:
is your app supposed to be run on x11 systems?

PR #828, opened 2026-01-02
added exception for io.github.tobagin.karere to access xdg-config/autostart

tobagin:
Startup manager needs direct access to manager autostart entries


bbhtt:
karere is a whatsapp client what do you mean by "startup manager". An app can use the portal to install autostart entries. Why does it need full access to autostart folder?

tobagin:
ok I will look into how to use the portal to do so, I sincerely thought it could only with the exception

PR #829, opened 2026-01-04
Update exceptions.json for org.kiwix.desktop

kiwixbot:
This is necessary to allow the new release `2.5.0` of Kiwix to pass the linting. See https://github.com/flathub/org.kiwix.desktop/pull/16


hfiguiere:
why aren't you using file portals? Using Qt brings this automatically.

bbhtt:
Is this still needed? Did you manage to investigate?

kelson42:
Honestly, all of this is really getting complicated to me. It seems to me now, I'm tracing a flatpak bug.

I have no clue about all these intrasecs flatpak details and I will handle that when I have more time, which means probably when we will do a new release.

bbhtt:
you can reopen once you have figured things out. this needs migrating to the new exceptions json structure anyways (it's in the docs)

PR #831, opened 2026-01-05
Add exception for ai.opencode.opencode

NN708:
The bundled `opencode` CLI binary requires host filesystem access to read and write project files.

For flathub/flathub#7489


bbhtt:
What paths does this realistically need to look for projects? Projects are not going to be stored in eg. `/usr`. 

> The bundled opencode CLI binary requires host filesystem access to read and write project files.

Why is this about a CLI? Does the GUI app need it or not?

PR #836, opened 2026-01-09
Add linter exception for io.github.sourcegit_scm.sourcegit (soon on flathub)

flexxxxer:
App provides ability to sign commits using gnupg, while the Flatpak's GPG needs to know which public keys are available - see https://wiki.gnupg.org/AgentForwarding.

App not yet on flathub, but will be soon.


flexxxxer:
Submission created.
As per https://docs.flathub.org/docs/for-app-authors/linter:

> Apps on Flathub or meant to be published on Flathub can apply for an exception by submitting a [pull request](https://github.com/flathub/flatpak-builder-lint/pulls) to the [exception file](https://github.com/flathub/flatpak-builder-lint/blob/master/flatpak_builder_lint/staticfiles/exceptions.json).

So I was thinking adding exceptions needed before app submission or at the same time with submission.

PR #838, opened 2026-01-13
Add exception for com.infinipaint.infinipaint to read/write to home directory

ErrorAtLine0:
Some users have run into issues saving to directories outside of Desktop/Documents. I am aware that these permissions can be changed using something like Flatseal, but if possible I'd like to have permissions for the home directory by default so that it could lead to less confusion for users who are unaware of Flatpak permissions.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1,4 +1,7 @@
 {
+    "com.infinipaint.infinipaint": {
+        "finish-args-home-filesystem-access": "Needed to save project files outside of standard directories."
+    },
     "org.vanillaos.ApxGUI": {
         "finish-args-flatpak-spawn-access": "Needed to execute Distrobox commands on the host"
     },


bbhtt:
why does it not use portals and save through a filechooser?

ErrorAtLine0:
I'm using the SDL3 library's native file picker to save and open files. Saving a file using the native file picker that SDL3 opens in a directory the flatpak doesn't have permissions for does still end up saving a file in that directory, but it saves it under a different name that's hidden to the user (something like ".xdp-FILENAME.infpnt-7uWOh5"). Is there anything else I would have to do in this case, or is it a bug in the library itself?

In any case, I'd still prefer having this exception, because there is a choice between the native file picker from SDL3 and the in program file picker, which can't save the file at all (outside of Desktop/Documents).

ErrorAtLine0:
Apparently this was an issue with file extensions. The application appended the file extension instead of the file picker, which meant that the app didn't have permissions to save at the new path with the appended extension. Making the native file picker decide the extension resolved that issue.

This still doesn't resolve the non-native file picker not being able to get permissions, and also not being able to drag-drop images onto the canvas from places the app doesn't have permissions for. This includes being able to drag/drop project files to open them instead of opening them from the file picker.

PR #840, opened 2026-01-14
haruna: add trash exception

g-fb:


g-fb:
I'm using KIO to delete files and that doesn't seem to support it.

bbhtt:
would be useful to file an upstream issue about it.

PR #877, opened 2026-01-30
Allow io.github.strikerx3.ymir to access host read only and home

Citrodata:
finish-args-host-ro-filesystem-access
This is used by ymir to allow users to load split bin files which on some systems, for example like the SteamDeck were the roms are on a SD Cards and not mounted in the users home folder.

finish-args-home-filesystem-access
This is wanted by the maintainer of the project to allow users to save custom configurations and later load them on the fly. These are not used by the config folder and more to be shared in the community.

Merge is in Response to this Pull by the maintainer 
https://github.com/flathub/io.github.strikerx3.ymir/pull/10

Issue was documented by Maintainer in Repository already
https://github.com/StrikerX3/Ymir/commit/f9d3ae80915ac77dd009cebb492cb073de51e2f2
Most comments about users wanting ro on host on discord by steamdeck users.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -6173,6 +6173,6 @@
     },
     "io.github.strikerx3.ymir": {
         "finish-args-host-ro-filesystem-access": "Required to load Roms correctly which where split into multiple bin tracks.",
-        "finish-args-home-filesystem-access": "Required in order to allow writing config files and custom files for loading of Sega Saturn IPL's."
+        "finish-args-home-filesystem-access": "Required in order to allow writing config files and custom files for loading of Sega Saturn IPL's. Also needed to load external game files and saved games folders"
     }
 }


bbhtt:
> when not run under Flatpak. 

why do you need the host location? If it is running as a flatpak it should save relative to `XDG_DATA_HOME=~/.var/app/$appid/data/`

bbhtt:
As far as I can see no exception is needed  here

> finish-args-host-ro-filesystem-access
This is used by ymir to allow users to load split bin files which on some systems, for example like the SteamDeck were the roms are on a SD Cards and not mounted in the users home folder.

Replace this by `/media, /mnt, /run/media` read-only access.

> finish-args-home-filesystem-access
This is wanted by the maintainer of the project to allow users to save custom configurations and later load them on the fly. These are not used by the config folder and more to be shared in the community.

This should not write to host's XDG_DATA_HOME. If you want to save configs, they should go inside `XDG_DATA_HOME` of the sandbox which is `~/.var/app/$appid/data/`

StrikerX3:
I'm still learning how the Flatpak sandbox works and what environment variables are made available. I'm not the author of the Flatpak release and did not develop Ymir with Flatpak in mind, so all of this is new to me. I'm working on adapting a few things here and there to better conform to the sandbox model and figuring out what permissions it really needs, so thanks for the pointers.

With that said, I'd like to clarify some details about how the app works.

> I have seen it and there is absolutely no reason it should need to write to host's XDG_DATA_HOME.

> This should not write to host's XDG_DATA_HOME. If you want to save configs, they should go inside `XDG_DATA_HOME` of the sandbox which is `~/.var/app/$appid/data/`

Was it the user home write permission that gave you the notion that Ymir needs to write to `$XDG_DATA_HOME` or anywhere else? I agree it's unnecessary now that I know the default permissions granted to apps are sufficient.

I have changed the app so that when it detects it's running under Flatpak it will not let the user pick any other location to write it configuration files to. It will always be stored in `$XDG_DATA_HOME/StrikerX3/Ymir`, as per the "Installed" mode.

With that, the only write accesses Ymir needs is:
- create the `$XDG_DATA_HOME/StrikerX3` directory
- create the `$XDG_DATA_HOME/StrikerX3/Ymir` directory
- create, modify and delete files and directories under `$XDG_DATA_HOME/StrikerX3/Ymir`

which, as I understand, is already granted by default.

Besides that, it needs read-only access to other parts of the filesystem where users store their game disc images. Those can be granted manually, and I think I provided enough directions for users to deal with that without having to rely on default permissions.

There is a command-line argument you can pass to force Ymir to store its configuration and state files anywhere in the filesystem, but this is an advanced feature that the majority of users don't know about, let alone use, and those that do tend to be more tech-savvy or developers that know they should be granting write permissions to the directories anyway (and carefully choosing where to store those files).

> As far as I can see no exception is needed here

I don't think the exceptions are needed after all, other than for the `/media, /mnt, /run/media` read-only accesses. I'll open a new PR with just those permissions.

PR #897, opened 2026-02-13
Add X11 and Wayland socket exception for RustDesk

fufesou:
Add `finish-args-contains-both-x11-and-wayland` exception for `com.rustdesk.RustDesk`. 

RustDesk requires the X11 socket for cursor image capture on the controlled side, and the Wayland socket for native Wayland protocol support (e.g. system shortcut inhibition).

See: https://github.com/rustdesk/rustdesk/pull/14302

This PR is blocked https://github.com/flathub/com.rustdesk.RustDesk/pull/31

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -4478,7 +4478,8 @@
         "finish-args-host-ro-filesystem-access": "Predates the linter rule"
     },
     "com.rustdesk.RustDesk": {
-        "finish-args-home-filesystem-access": "Predates the linter rule"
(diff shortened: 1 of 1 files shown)


bbhtt:
The exception won't be given.

yakushabb:
You can use wayland + fallback-x11

fufesou:
Thanks for the clarification.

However, we do need `--socket=x11`; otherwise, the cursor image capture functionality will not work.

We previously used `wayland + fallback-x11`, but had to change to `--socket=x11` specifically because of this requirement.

https://github.com/rustdesk/rustdesk/pull/13544

PR #905, opened 2026-02-17
Add exception for appid-url-not-reachable on io.github.jalaucapstones.pacman-recreation

CodeWithBotinaOficial:
Add exception for appid-url-not-reachable on io.github.jalaucapstones.pacman-recreation

The linter derives https://github.com/jalaucapstones/pacman-recreation which 404s, but the real repo is https://github.com/JalaU-Capstones/pacman-recreation (capital letters and hyphen in username). The domain is fully controlled by me (Diego Alejandro / Code With Botina), so this is a false positive.

Thanks!


hfiguiere:
Same as #904 

this exception will not be granted. It is documented how app-id are supposed to be made.

PR #906, opened 2026-02-18
Restore exception for AbiWord

hfiguiere:
It was automatically removed in commit 67ebcaece4aa9cbf64633a209ab7b12d7551fc05 erroneously

See https://github.com/flathub/com.abisource.AbiWord/pull/49
and
https://github.com/flathub/com.abisource.AbiWord/pull/50


bbhtt:
Seems like it's because the beta branch doesn't have `home` (it has `xdg-documents`) while master has `home`.

hfiguiere:
The beta branch will get portal support. It's on my todo list.

PR #911, opened 2026-02-20
added host access to org.cubocore.CoreFM

rahmanshaber:
org.cubocore.CoreFM is a file manager and requires access to all the file system to able to browse all files.

only access to the run, mmt is not enough, as file manager it brakes the usability like viewing, making backup system files.


rahmanshaber:
I totally agree with you. I'm pushing all of our app here because of reach flatpak has. I hope you guys think about this part also. 

>> Unless these applications are being officially submitted by the upstream developers and/or upstream has explicitly committed to supporting Flatpak they will not be accepted.

I think this app fulfils both of this. we made lot of faltpak related changes in v5.

anyway, thanks for the great work you guys are doing, @bbhtt.

PR #938, opened 2026-03-13
Add more exceptions for io.github.sockeye_d.godl

sockeye-d:
Godot requires these to function, I wasn't properly requesting those permissions before.


sockeye-d:
Both options are viable, there's no reason to artificially limit users to only X11

bbhtt:
I won't be merging this with that

sockeye-d:
Wayland is considered an experimental platform. However it works better already for many cases such as computers with high-precision touchpad input devices. As such, it defaults to X11, because it's considered more "fully featured". However, the user should be able to set it to use Wayland, as they would with a native Godot installation.

In addition to this, _Godot itself_ has these permissions, which is why if I want to be able to properly launch it, _I_ need those permission too.

bbhtt:
Godot has it because they wanted to test the experimental wayland in godot's beta branch at the time of making the exception. The beta branch is meant for testing and not for long term consumption by end users. The exceptions don't differentiate between beta and stable branches.

While you want this for your app's stable branch. So still no

If it is experimental people can use it with `flatpak-override` there is no reason to have the permission by default.

bbhtt:
See the PR description https://github.com/flathub-infra/flatpak-builder-lint/pull/433

bbhtt:
Looks like I misread when accepting Godot's exceptions, they shouldn't have been accepted for stable branches.

sockeye-d:
Regardless, I think my point still stands. I can't use just x11 because Godot has an option to enabled Wayland support. I can't use wayland and fallback-x11, because Godot launches by default with an x11 session and will fail to launch on Wayland systems. As such, for Godot to have all of its features, I need to have both X11 and Wayland socket access.

sockeye-d:
It's not a command line argument (although it also is), it's an option in the UI
Screenshot_20260314_001242

sockeye-d:
I'm not; the flatpak-spawn permission is for Godot to flatpak-spawn programs like Blender. Godot is already a self-contained program that discovers dependencies at runtime, and works across a wide variety of distros with minimal dependencies. I also think that it's a lot more intuitive to the user if Godot can't find .NET/JVM in the host, where they can control it, rather than in the sandbox, where they have no control over it. Moving those dependencies onto the user out of the flatpak also makes it smaller.

bbhtt:
Ok then

bbhtt:
> Would it be better to just flatpak-spawn Godot directly, instead of requesting its permissions?

Was this done? Is the exception needed still?

PR #976, opened 2026-04-09
Add exception for org.cvfosammmm.Lemma

cvfosammmm:
Required for importing markdown files with images and attachments, as only the main file can be accessed through the portal.


bbhtt:
this should be home plus directories you need, i'm not sure if anyone is referencing files inside /usr or /etc and even if they are they won't appear at the same location inside the sandbox, so the markdown file will be broken anyways.

even better if it can be scoped to some selected subdirs of home.

PR #979, opened 2026-04-14
Add exception for io.github.rafaelfassi.QLogExplorer

rafaelfassi:
I have received some complaints that QLogExplorer is unable to open certain log files.
The reason is because those logs are not inside /var/log which is the only location outside home QLogExplorer has filesystem permission to read.

Different distros and applications may store logs in a variety different locations, and there is no way to predict all possible places the log files can be stored. Restricting the access to specific locations only makes the software less useful for developers, devops and system admins who are the target users for this application.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -4541,8 +4541,7 @@
     },
     "io.github.rafaelfassi.QLogExplorer": {
         "stable": {
-            "finish-args-home-filesystem-access": "Predates the linter rule",
-            "finish-args-host-var-access": "Predates the linter rule"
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -4541,8 +4541,7 @@
     },
     "io.github.rafaelfassi.QLogExplorer": {
         "stable": {
-            "finish-args-home-filesystem-access": "Predates the linter rule",
-            "finish-args-host-var-access": "Predates the linter rule"
+            "finish-args-host-ro-filesystem-access": "Predates the linter rule"
```
Erick555: you may put actual reason for exception since this doesn't predate linter anymore

rafaelfassi: Thanks for the review.
Description updated.

PR #980, opened 2026-04-15
Add exception for fr.arnaudmichel.launcherstudio

MrArnaudMichel:
Add network acces because it's required to search and download Lucide icons from the Iconify API when the user customizes their launchers in the new release of my app.


CodedOre:
There is currently no lint against network access. So you don't need an exception for that either.

MrArnaudMichel:
> There is currently no lint against network access. So you don't need an exception for that either.

Ok perfect !

PR #982, opened 2026-04-16
Add exceptions for io.github.forkgram.tdesktop

ilya-fedin:


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -3967,6 +3967,14 @@
             "finish-args-home-filesystem-access": "Predates the linter rule"
         }
     },
+    "io.github.forkgram.tdesktop": {
+        "*": {
+            "flathub-json-automerge-enabled": "Telegram enforces client version cutoffs; stale builds lose server connectivity",
+            "module-qt-checker-tracks-commits": "Manually updated",
+            "module-tde2e-checker-tracks-commits": "Manually updated",
+            "module-tg_owt-checker-tracks-commits": "Manually updated"
```
bbhtt: how are these going to be "manually updated"? I don't see any mechanism for it in the PR and with the x-checker paras the global one will just update them.

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -3967,6 +3967,14 @@
             "finish-args-home-filesystem-access": "Predates the linter rule"
         }
     },
+    "io.github.forkgram.tdesktop": {
+        "*": {
```
bbhtt: this is going to stable only? why do you need beta?

ilya-fedin: oops, copy-pasted

flatpak_builder_lint/staticfiles/exceptions.json
ilya-fedin: It checks upstream file with the deps commits/tags which is manually updated upstream

bbhtt: by that definition every release of every software is "manual"

"updating manually" here means you update the commit in the manifest manually and opt out of the global x-checker action.

ilya-fedin: But that wording has worked previously? It was initially introduced by you, in 082d00daeb8c6470d9cc4b53a3e625ff3fe1b652... Which wording should be used now then?

bbhtt: that was grandfathered and iirc the repos were using a custom GH action at that time instead of the global checker workflow.

bbhtt: also iirc they were updated in beta first and then later merged to stable, telegram has a unique workflow I don't fully remember the state it was in a year ago.

ilya-fedin: > that was grandfathered and iirc the repos were using a custom GH action at that time instead of the global checker workflow.

The exception for updating via custom GH action is `external-gitmodule-url-found` which you said couldn't be given. The `module-tg_owt-checker-tracks-commits` (`module-tg_owt-checker-uses-commit-query` in your commit) always was updated via x-checker-data from upstream deps file and that's where you used `Manually updated`. Do you want the description just to be re-worded or is it unacceptable anymore?

> also iirc they were updated in beta first and then later merged to stable, telegram has a unique workflow I don't fully remember the state it was in a year ago.

The workflow is that if upstream version is a beta one, the change gets merged only to beta. If it's a stable one, it's merged to both beta and master (with the help of a GH action). Forkgram follows only stable releases so this complexity isn't needed.

bbhtt: > where you used `Manually updated`

I also used "grandfathered" and yes this is not "manually updating". It is not acceptable to get the exception for new apps.

It's not possible for me to remember specifics of every 3500 apps over several years when adding exceptions. It's why I keep saying to ignore old exceptions and you should generally assume that those are being grandfathered if I am adding exceptions along with new checks.

ilya-fedin: > It is not acceptable to get the exception for new apps.

Can you explain why at least? The commits for these deps already have to be updated in two files upstream and having a third one in another repo isn't really appealing...

bbhtt: we don't allow tracking commits because there's always the possibility of upstream committing too often and each one generating an update PR and exhausting the build capacity. then yours and everyone else's PRs will get delayed. there is no unlimited build capacity.

the way to get out of it is you 

a) [opt out of global x-checker run](https://docs.flathub.org/docs/for-app-authors/external-data-checker#disable)
b) update manually or use your custom workflow that runs on some reasonable interval (probably synced with when updates are pushed out, you can even trigger it from external repo to make sure the PR gets created almost immediately as updates are pushed in upstream repo)

ilya-fedin: > we don't allow tracking commits because there's always the possibility of upstream committing too often and each one generating an update PR and exhausting the build capacity. then yours and everyone else's PRs will get delayed. there is no unlimited build capacity.

But it uses the commit specified in the tdesktop module (the app itself) to fetch the upstream deps file. It won't get commit changes unless the app itself is updated (i.e. tde2e, tg_owt and Qt patches all get updated only with the app itself which will trigger a build anyway).

> [opt out of global x-checker run](https://docs.flathub.org/docs/for-app-authors/external-data-checker#disable)

You mean having own instance of flathub-external-data-checker?

bbhtt: > It won't get commit changes unless the app itself is updated (i.e. tde2e, tg_owt and Qt patches all get updated only with the app itself which will trigger a build anyway).

ok that sounds reasonable, you can add that as a reason.

PR #983, opened 2026-04-16
Update exceptions for org.telegram.desktop

ilya-fedin:
https://github.com/flathub/org.telegram.desktop/pull/1063

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -8397,8 +8397,8 @@
     "org.telegram.desktop": {
         "*": {
             "appid-ends-with-lowercase-desktop": "app-id predates this linter rule",
-            "external-gitmodule-url-found": "Predates the linter rule",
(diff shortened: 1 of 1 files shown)


bbhtt:
same question as the other how is this getting manually updated?

ilya-fedin:
It checks upstream file with the deps commits/tags which is manually updated upstream

ilya-fedin:
From https://github.com/flathub/flathub/pull/8183 discussion I assumed it's better to handle patches via x-checker-data than via git module updated via a custom github action. It also solves the problem of Qt updating before patches are updated resulting in a useless failing build. If the change is not welcomed, just say it straight to not to waste both my and your time on this.

bbhtt:
was this being updated through app's tag as well? you can reopen and update the reason as well, same as https://github.com/flathub-infra/flatpak-builder-lint/pull/982

ilya-fedin:
Yeah, github doesn't show open button though, I assume only you can reopen

PR #985, opened 2026-04-17
Add exceptions for de.wwwtech.gitte

ckruse:
I was told to request an exception: https://github.com/flathub/flathub/pull/8387#discussion_r3095904811

- `--filesystem=home`: i need access to the git repositories the user has configured, as well as access to ~/.git* files to respect the users git configuration. Also the user is able to configure to use a specific SSH key per repo and remote for remote operations, so access to ~/.ssh is also needed
- `--socket=ssh-auth`: needed for using the SSH transport for remote git operations

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -2940,6 +2940,12 @@
             "finish-args-home-filesystem-access": "Predates the linter rule"
         }
     },
+    "de.wwwtech.gitte": {
+        "stable": {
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -2940,6 +2940,12 @@
             "finish-args-home-filesystem-access": "Predates the linter rule"
         }
     },
+    "de.wwwtech.gitte": {
+      "*": {
```
bbhtt: "stable"

ckruse: oh, sorry. Fixed.

PR #991, opened 2026-04-22
Adding exception for io.github.intoolswetrust.JSignPdf

Sesivany:
I'm requesting an exception based on the suggestion in the admission review: https://github.com/flathub/flathub/pull/8338#discussion_r3065791644


bbhtt:
Why can't this be limited to xdg-documents/downloads etc.?

Sesivany:
I think it would probably work for accessing the documents to sign, but JSignPdf also needs access to keystores with signing certificates, these are typically stored in other locations and I don't think it's wise to ask the user to move them around, so the app can see them.

PR #995, opened 2026-04-24
io.github.moonlight_mod.moonlight-installer: add new discord installation paths

slonkazoid:
discord recently turned the main app directory into a shim that downloads and calls ~/.config/(installation)/app-(version)/(binary), so we will need to include those new paths


Erick555:
I think it would be helpful to write in the title and commit for which app those rules are so you don't have to look into diff to know that.

bbhtt:
How would I know these are the correct paths used by Discord other than installing it? Did they announce it somewhere or is the information available somewhere?

slonkazoid:
> How would I know these are the correct paths used by Discord other than installing it? Did they announce it somewhere or is the information available somewhere?

the `discord-canary` shell script in https://canary.discord.com/api/download/canary?platform=linux&format=tar.gz (2.1 MB)

PR #1000, opened 2026-04-28
Adding exception for gov.nasa.giss.Panoply

Mailaender:
Required for https://github.com/flathub/flathub/pull/8499. 

grafik

The application uses Java Swing, which starts in `$HOME` and will otherwise be broken.


bbhtt:
you have drop the modification of chirp.

Mailaender:
Added back the typo.

PR #1004, opened 2026-05-01
Add X11 + Wayland exception for Jetbrains Rider

TonimatasDEV:
It's needed for https://github.com/flathub/com.jetbrains.Rider/pull/169 to fix the issue 168 https://github.com/flathub/com.jetbrains.Rider/issues/168

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1985,7 +1985,8 @@
         "stable": {
             "finish-args-arbitrary-dbus-access": "IDEs tend to require direct dbus access",
             "finish-args-flatpak-spawn-access": "the app predates this linter rule",
-            "finish-args-host-filesystem-access": "Predates the linter rule"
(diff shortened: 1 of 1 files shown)


bbhtt:
No in this case it should use x11 socket only and the "use wayland by default" commit needs to be reverted.

flexxxxer:
> No in this case it should use x11 socket only and the "use wayland by default" commit needs to be reverted.

Why "use x11 socket only"? I wanna benefit from features and experience that JetBrains Rider provides running under Wayland while be able to utilize x11 socket for apps development for platforms JetBrains Rider was made for.

Experience with JetBrains IDEs running under Wayland just much better: fonts, multi-window experience, etc.

TonimatasDEV:
> No in this case it should use x11 socket only and the "use wayland by default" commit needs to be reverted.

JetBrains now uses Wayland by default and they are actively working to improve it. Some IntelliJ-based IDEs have the same issue so people will need X11 or Wayland depending on their development stack. We should support both environments in some IDEs (IntelliJ, CLion, Rider...) to be sure they can use any library.

PR #1005, opened 2026-05-01
Add exception for io.github.fidele007.akkhara

ghost:
> "finish-args-own-name-com.fidele007.akkhara": "The app requires this access to run"

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3964,6 +3964,11 @@
             "finish-args-host-etc-filesystem-access": "Predates the linter rule"
         }
     },
+    "io.github.fidele007.akkhara": {
+        "stable": {
+            "finish-args-own-name-com.fidele007.akkhara": "The app requires this access to run"
+        }
+    },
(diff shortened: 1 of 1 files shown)


bbhtt:
This is not sufficient explanation and I have no intention of granting this

bbhtt:
You should figure out the exact permissions needed https://docs.flatpak.org/en/latest/sandbox-permissions.html#d-bus-access or form a better explanation than "it just needs this".

ghost:
I updated the PR to the exact permission. Thanks for guiding.

bbhtt:
In this case the question is why does it not follow the Flatpak ID? Is the app maintained by upstream, or are they active? This is something that should be fixed in the app.

ghost:
@fidele007, this app (io.github.fidele007.akkhara) appears to be validated by you.

Would you have time to check this issue? Thanks.

ghost:
No answer from the developer. Can we proceed now?

bbhtt:
Unlikely as I said it should be fixed in the app. If the developer is not responsive, it should be EOLed.

Let's close this.

PR #1008, opened 2026-05-03
Requesting exceptions for io.github.arijanj.Mimic

ArijanJ:
Hello, Mimic is an application that lets users graphically set their default applications (mimeapps.list) regardless of which DE/WM they're using; even if their DE has a 'default applications' menu, Mimic gives much more granular control. (https://github.com/flathub/flathub/pull/8543)

For that purpose, it needs to iterate through all .desktop files on the system. I've added the most common locations, and this works fine on a bunch of different distributions I've tried. From all the research I've done, these should be the minimum permissions possible to put the app in a working state.

I understand this is a bit of a dubious use-case, but I think having something this available on Flathub would be very useful, as I use the somewhat similar https://flathub.org/en/apps/io.github.fabrialberio.pinapp app on all my devices.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3720,6 +3720,16 @@
             "finish-args-home-filesystem-access": "Predates the linter rule"
         }
     },
+    "io.github.arijanj.Mimic": {
+        "stable": {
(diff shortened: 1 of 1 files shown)


bbhtt:
These are not correct it has :ro not -ro you have to copy verbatim

ArijanJ:
Ah sorry, I saw all of the older entries using -ro so I thought I had to match them

bbhtt:
json is invalid.

ArijanJ:
Fixed

PR #1015, opened 2026-05-06
finish-args: Drop device=input and usb lints

bbhtt:
Enough time has passed since 1.16.0 and a new LTS release of Ubuntu and soon a new stable release of Flatpak will be available. 1.14.x which didn't had this is unsupported entirely, so let's allow these.

After this, `--device=input, usb` can be used as long as `--require-version=X` (X >=1.16.0) is used. The lint still requiring a `--require-version=X is kept as is.

### DIFF
--- a/flatpak_builder_lint/checks/finish_args.py
+++ b/flatpak_builder_lint/checks/finish_args.py
@@ -143,20 +143,6 @@ class FinishArgsCheck(Check):
                         "finish-args-insufficient-required-flatpak: finish-args has"
                         + " 'input' or 'usb' device but 'require-version' is not >=1.16.0"
                     )
-                if dev == "input":
-                    self.errors.add("finish-args-has-dev-input")
-                    self.info.add(
-                        "finish-args-has-dev-input: This permissions is not backwards"
-                        + " compatible with a supported Flatpak release 1.14.x and"
-                        + " requires an exception"
-                    )
-                if dev == "usb":
-                    self.errors.add("finish-args-has-dev-usb")
-                    self.info.add(
-                        "finish-args-has-dev-input: This permissions is not backwards"
-                        + " compatible with a supported Flatpak release 1.14.x and"
-                        + " requires an exception"
-                    )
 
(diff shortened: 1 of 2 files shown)


bbhtt:
Planning to merge this when Flatpak 1.18.0 is out

electricbrass:
Why does this require `--require-version`? According to the Flatpak [documentation](https://docs.flatpak.org/en/latest/sandbox-permissions.html#conditional-permissions), it is possible to use these in a backwards-compatible way with conditional permissions:

> 1. The unconditional grant `--socket=x11` can be tightened using conditional permissions to `--socket-if=x11:!has-wayland`. This allows access to X11 only when a Wayland desktop session is not available. **To preserve backwards compatibility, the following pattern can be used**:

```
--socket=x11
--socket-if=x11:!has-wayland
```

> **This allows older Flatpak versions which do not understand the conditional permissions function** by allowing X11 access always while newer Flatpak, which understands the conditional system will allow X11 access only when the session is not Wayland.

> 2. If an application requires only access to input device permission, the following flags can be used to move away from `--device=all` to `--device=input`:

```
--device=all
--device-if=all:!has-input-device
--device=input
```

> **This allows older Flatpak versions which do not understand the input device permission to function** by having the broader all access. Newer Flatpak versions which understand the conditional system (and therefore understands the input permission) will deny all due to --device-if=all:!has-input-device and allow only input due to --device=input.

It seems like it would be better if the linter allowed not using `require-version` as long as the permissions were set up in a way that still provided `--device=all` for older versions, like in these examples.

(This is assuming Flathub is building with 1.17/1.18)

bbhtt:
Conditional permissions aren't in a stable release and has not been deployed to Flathub. It's not going to be deployed until some months after a stable release.

A plain `--device=input, usb` is not conditional and is not backwards compatible. If someone wants to use that it will require specifying a require-version.

Once conditional is in a stable release and has been deployed to Flathub, someone can of course choose to use that syntax and enable a backwards compatible behaviour. Then require version will not be needed.

PR #1016, opened 2026-05-06
Add exception for io.github.kyuyrii.meganimus

Kyuyrii:
The reason you need "finish-args-home-filesystem-access" and "finish-args-flatpak-appdata-folder-access" is that without this access, when selecting a file, the path obtained is not the actual path of the native game, .exe file, or ROM, and the Document Portal isolates the selected file.

So if the user selects something with the Document Portal, the .desktop file created by Meganimus will break.

If it's a native game, the file that runs the game will not find the game files.

If it's an .exe file, for example, if it's to be run with the Flatpak version of Wine (org.winehq.Wine), using a Document Portal path will prevent Wine from executing the .exe file. This is because even if it's inside the `$HOME/.var/app/org.winehq.Wine` folder, Wine won't be able to access that path created by the Document Portal. Even if it could, the .exe file would be isolated, causing the same problem that occurs with native games.

https://github.com/user-attachments/assets/ddbeef22-a445-480d-8a90-66b85f26d22e

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -4903,6 +4903,12 @@
             "finish-args-host-tmp-access": "Predates the linter rule"
         }
     },
+    "io.github.kyuyrii.meganimus": {
+        "stable": {
+            "finish-args-home-filesystem-access": "Required for .desktop files to use the actual file path selected by the user.",
+            "finish-args-flatpak-appdata-folder-access": "Required for .desktop files to use the actual file path selected by the user."
+        }
+    },
     "io.github.kyuyrii.zordeer": {
         "stable": {
             "finish-args-home-filesystem-access": "The UMU launcher needs direct access to function. Using the Document Portal prevents it from working."


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -4903,6 +4903,12 @@
             "finish-args-host-tmp-access": "Predates the linter rule"
         }
     },
+    "io.github.kyuyrii.meganimus": {
+        "stable": {
+            "finish-args-home-filesystem-access": "Required for .desktop files to use the actual file path selected by the user.",
```
bbhtt, line 4908: actual path to what exactly?

bbhtt: also the portal has an API to get actual host paths.

Kyuyrii: > actual path to what exactly?

The path the user selects, the Document Portal changes the path the user selected, and also isolates the selected file. In the video above I show exactly the problem that the Document Portal creates.

To solve this, I would have to be able to disable the Document Portal, except when the user selects an image file, which is the only type of file that Meganimus actually needs to access.

In all other cases, the important thing is only to receive the real file path.

bbhtt: That's exactly where you are supposed to use GetHostPaths and get the actual path on host from the portal.

Exception will not be granted for this, the feature exists, you can use it.

Kyuyrii: But how do I use GetHostPaths? I've never heard of it.

Kyuyrii: How can I follow your wishes if you closed the pull request and don't want me to open another one?

You may not want to deal with me, but there are other reviewers.

bbhtt: > But how do I use GetHostPaths? I've never heard of it.

https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Documents.html#org-freedesktop-portal-documents-gethostpaths

https://github.com/search?q=%22GetHostPaths%22+language%3Apython&type=code

> but there are other reviewers.

This isn't a gift shop, you don't get to choose and you are doing the same thing here as you did in the submission PR so locked for my peace

Allexio:
@Kyuyrii I don't know if you know this but there are ways to resolve flatpak paths to "normal" ones.
In my app my workflow is:
1. User selects path with file picker
2. I get a "flatpak-style" path
3. I translate that path to a standard non-flatpak path
4. I ask the user to provide access to that path via a console command (and restart)
5. I now have proper access to that path

That being said - if you're looking to get paths to games there are better ways.
Steam and Heroic _both_ have a record of where the libraries are stored. So you don't actually need to translate those paths, you just need to read the config files to get the full "real" paths you need and then ask the user to provide access to those.

This is exactly what I do in my app, feel free to check out my code (it's in python).

PR #1019, opened 2026-05-08
Add exceptions for io.github.ForWard_Technologies_LLC.Pylux

forward-technologies:
Add linter exceptions for `io.github.ForWard_Technologies_LLC.Pylux` (Flathub submission: flathub/flathub#8607).

- `finish-args-host-os-ro-filesystem-access`: Requires read-only access to host-os to load Vulkan GPU drivers from the host filesystem, necessary for hardware-accelerated video decoding on Steam Deck and other Linux systems.
- `finish-args-login1-system-talk-name`: Requires `org.freedesktop.login1` to inhibit system sleep and screensaver during active remote play streaming sessions.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -4707,6 +4707,12 @@
             "finish-args-login1-system-talk-name": "Predates the linter rule"
         }
     },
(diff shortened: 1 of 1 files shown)


bbhtt:
GPU drivers are in the extensions, you cannot just use whatever you want from host like that.

There is a portal for the second, so you should use that.

PR #1020, opened 2026-05-08
Update exceptions.json

CodeMouse92:
Timecard is already designed to store config and log files in the home folder, and users will expect to find them there. I've been doing this since before the linter blocked it, and I'm trying to update the package.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -537,7 +537,6 @@
             "finish-args-host-filesystem-access": "Predates the linter rule"
         }
     },
-    {
     "com.codemouse92.timecard": {
         "*": {
             "finish-args-host-filesystem-access": "Predates the linter rule"
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -538,7 +538,7 @@
         }
     },
     {
-  "com.codemouse92.timecard": {
+    "com.codemouse92.timecard": {
         "*": {
             "finish-args-host-filesystem-access": "Predates the linter rule"
         }
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -538,8 +538,8 @@
         }
     },
     "com.codemouse92.timecard": {
-        "*": {
-            "finish-args-host-filesystem-access": "Predates the linter rule"
+        "stable": {
+            "finish-args-host-filesystem-access": "required to allow the user to store and retrieve logs and configuration files in existing locations in their home directory"
         }
     },
     "com.core447.StreamController": {
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -537,6 +537,13 @@
             "finish-args-host-filesystem-access": "Predates the linter rule"
         }
     },
+    {
+  "com.codemouse92.timecard": {
+        "*": {
+            "finish-args-host-filesystem-access": "Predates the linter rule"
+        }
+    },
+    },
     "com.core447.StreamController": {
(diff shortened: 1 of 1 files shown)


bbhtt:
You must have discovered time travel.

CodeMouse92:
@bbhtt I mean, time lord, TARDIS, sure, but I don't know beyond that. I clicked the edit icon on the file here on GitHub, let it create the fork, made my one entry change, and opened the PR. But we all know GitHub has been weird lately.

As to the exception, I've been using this permission for Timecard since long before the linter rule blocked it.

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -537,6 +537,11 @@
             "finish-args-host-filesystem-access": "Predates the linter rule"
         }
     },
+    "com.codemouse92.timecard": {
+        "*": {
+            "finish-args-host-filesystem-access": "Predates the linter rule"
```
Erick555: I see your app has [home](https://github.com/flathub/com.codemouse92.timecard/blob/master/com.codemouse92.timecard.yaml#L18) not host access and the [error](https://github.com/flathub-infra/vorarbeiter/actions/runs/25553676397/job/75007385956) clearly is:

> Error: **finish-args-home-filesystem-access** error found in linter manifest check.

You should also provide the actual reason instead of boilerplate line

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -537,6 +537,11 @@
             "finish-args-host-filesystem-access": "Predates the linter rule"
         }
     },
+    "com.codemouse92.timecard": {
+        "*": {
```
Erick555: use `stable` instead of `*`

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -537,6 +537,11 @@
             "finish-args-host-filesystem-access": "Predates the linter rule"
         }
     },
+    "com.codemouse92.timecard": {
+        "stable": {
+            "finish-args-home-filesystem-access": "required to allow the user to store and retrieve logs and configuration files in existing locations in their home directory"
```
bbhtt, line 542: why does a timetracker app need to store and retrieve logs? what are these logs and where are they being saved / by whom?

CodeMouse92: The _time logs_ of the _time being tracked_ by the app, which _the user_ will expect to be able to find and customize the save location of _in their home folder_. This is a core functionality of the app since long before the linter was added.

CodedOre: If you think about "logs" in the context of a software, you first think of logs about the operation of the software (think warnings, errors, etc.).

Not user-created data about time tracking. So, maybe this could be formulated more clearly. \
Or, even better, the FileChooser portal could be used when the user picks tge output directory. Then no him permission is needed. 

Apart from that:\
Does the application store configuration in the home folder? The XDG base directory specification exists for a reason and should be used.

CodeMouse92: I am updating a long-time-existing application in my sparest of spare time, I do not have the time to completely rearchitecture existing behavior to meet the changing expectations of a packaging format. So if that is going to be necessary, then I suppose I'll have to accept that Flatpak is no longer a viable distribution method.

Erick555: Your app was inactive for 4 years and get EOL status which was revoked just few days ago so it's basically like a brand new app on flathub right now.

CodeMouse92: k, well, if Flatpak would rather I just not use their platform, I'll just move on. I have higher priorities than rewriting a working application's actual implementation and user experience to meet the underdocumented expectations of a _packaging format_. Pity too, since I am revising the packaging chapter in the book, but there are other packaging formats I can explain instead. Cheers!

PR #1023, opened 2026-05-08
Use the same exceptions every branch of io.github.antimicrox.antimicrox

pktiuk:
I want to create beta branch for my app to start working on some issues

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3715,7 +3715,7 @@
     },
     "io.github.antimicrox.antimicrox": {
         "*": {
-            "finish-args-home-filesystem-access": "Predates the linter rule"
+            "finish-args-home-filesystem-access": "Required until https://github.com/flatpak/xdg-desktop-portal-gtk/issues/191 is fixed"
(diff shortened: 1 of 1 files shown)


bbhtt:
Update the reason too with the actual information on why it is needed.

pktiuk:
Done

bbhtt:
This is specific to using the Qt app on GNOME or GNOME/GTK portal?

pktiuk:
TBH I haven't checked this by myself, but some users claim that it is the case

https://github.com/flatpak/xdg-desktop-portal-gtk/issues/191#issuecomment-474259765

> I guess this is specific for GTK portal / Qt app combo.
>  Qt app with KDE portal appends the extension correctly, so does GTK app with GTK portal.

In my app this caused issue: https://github.com/AntiMicroX/antimicrox/issues/70 which point also at https://github.com/flatpak/xdg-desktop-portal/issues/496

bbhtt:
It's an old bug in Qt https://qt-project.atlassian.net/browse/QTBUG-91091

I see there was a fix posted recently I can merge the exception now but if it does get fixed somewhere you should try to drop or reduce the permission.

PR #1024, opened 2026-05-09
Add pylux exception

forward-technologies:
Add linter exception for `io.github.ForWard_Technologies_LLC.Pylux` (Flathub submission: flathub/flathub#8607).

**`finish-args-login1-system-talk-name`**: Required to receive the `PrepareForSleep` signal from logind so active streaming sessions can be cleanly torn down before the system suspends. Idle inhibit is already handled via `org.freedesktop.portal.Inhibit` (no manifest permission needed); `login1` is used solely for the `PrepareForSleep` signal, which has no portal equivalent. This is especially relevant on Steam Deck where suspend is the primary way to end a session.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3523,6 +3523,11 @@
             "finish-args-flatpak-appdata-folder-access": "Predates the linter rule"
         }
     },
+    "io.github.ForWard_Technologies_LLC.Pylux": {
+        "new-pr": {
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -3523,6 +3523,11 @@
             "finish-args-flatpak-appdata-folder-access": "Predates the linter rule"
         }
     },
+    "io.github.ForWard_Technologies_LLC.Pylux": {
+        "*": {
```
bbhtt:
```suggestion
        "stable": {
```

PR #1032, opened 2026-05-18
Add exception for gay.brie.CheesePaper

ByteOfBrie:
Based on flathub/flathub#8714

While most of the typical user interaction will be done entirely via UI, there are a few cases where users will need to interface with files directly, including both the `xdg-config/cheese-paper` and `xdg-data/cheese-paper` directories. These instances should be fairly rare, but they do exist, and are documented in the manual (and, to a lesser extent, in the UI).


bbhtt:
this won't be given flatpak apps are not supposed to use host xdg locations. if you need to interface you should browse the flatpak locations, not merge them with host.

PR #1033, opened 2026-05-18
Add exception for net.huangyuhui.hmcl

first-storm:
1. the correct domain is hmcl.huangyuhui.net or www.huangyuhui.net. And i have already asked the upstream dev to fix certificate issue.

2. Because JavaFX only supports X11 but some Minecraft build supports Wayland. If I choose to use fallback-x11 and wayland, HMCL won't start.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -5495,6 +5495,12 @@
             "finish-args-contains-both-x11-and-wayland": "Predates the linter rule"
         }
     },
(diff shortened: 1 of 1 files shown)


hfiguiere:
> 1. the correct domain is hmcl.huangyuhui.net or [www.huangyuhui.net](http://www.huangyuhui.net). And i have already asked the upstream dev to fix certificate issue.

then it will wait the certificate is fixed.

hfiguiere:
the use of copilot is strictly prohibited.

hfiguiere:
there won't be another warning.

PR #1035, opened 2026-05-20
Add exception for net.windower.Lumoria

rysas:
Adjusting permissions in Flathub Manifest to be more in line With [XIVLauncher](https://github.com/flathub/dev.goats.xivlauncher) and [Lutris](https://github.com/flathub/net.lutris.Lutris). After user testing, users are running into quite a few different issues with portals and Wine. Massive slowdowns due to overhead of the portal and more. Our users are moving from Lutris so these permissions will not come as a surprise.

In relation to: https://github.com/flathub/net.windower.Lumoria/pull/11

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -5793,6 +5793,11 @@
             "flathub-json-automerge-enabled": "Predates the linter rule"
         }
     },
+    "net.windower.Lumoria": {
+        "stable": {
+            "finish-args-home-filesystem-access": "Wine/Proton have issues running game under sandbox, occasionally settings get reset due to launcher bugs. SteamOS xdg-document-portal process takes up all CPU when installing and hangs device if outside of sandbox."
+        }
+    },
     "net.wz2100.wz2100": {
         "*": {
             "finish-args-contains-both-x11-and-wayland": "Predates the linter rule"


bbhtt:
You aren't going to get exceptions accepted for undocumented issues. Previous instances of similar apps having it is not an acceptable reason either.

> Massive slowdowns due to overhead of the portal and more.

Have these issues been reported to the components involved?

The rest of reasoning seems spread around "Wine has issues, proton has issues and SteamOS has issues". It's not very clear what exactly are the issues or if there have been attempts at solving it

rysas:
Thanks for the reply. Had limited time from the holiday weekend and just catching up. 

> You aren't going to get exceptions accepted for undocumented issues. Previous instances of similar apps having it is not an acceptable reason either.

Understood and fair. A few work arounds have been discovered by the community on some of our issues and I am working to implement those. Just not problems with the xdg-document-portal / Wine on the Steam Deck causing it to hang and require a forceful restart. 

The issue seems to be more so related to ngen.exe from installing .NET 4.8 / fsync in Wine over the xdg-documement-portal and not the portal itself. The hard part (joys) of Linux, figuring out what component of something is causing the root issue! For now we'll be providing instructions for users that want to install out of the XDG space for the app on SteamOS (flatseal/cli).

I couldn't find much related to this, so I will do more due diligence and see if I can figure out how to trace this thing down and report where necessary. That will take an amount of time, so for now perhaps we can close this issue and I will reopen if that avenue leads nowhere.

bbhtt:
If you find something it's better to open upstream issues first and link them here, then we have some confirmation of the problem and it may be fixed.

rysas:
Closing this now.

Workarounds for aforementioned problems have been found and it's not really the problem of Flatpak sandboxing.

Thank you for your patience. This has been a huge learning experience and Wine is a gnarly beast!

PR #1036, opened 2026-05-21
Add org.tahoma2d.Tahoma2D exception

manongjohn:
Submitting this exception to be added for org.tahoma2d.Tahoma2D to go along with my Flatpak app submission (https://github.com/flathub/flathub/pull/8742).

Tahoma2D is an animation software that is used regularly with other drawing software in an animation pipeline and needs to be able to share image files between them which could be anywhere on their system.

Edi: Also, Tahoma2D needs to be able to configure the location and run external explications, ffmpeg and rhubarb to ingest certain image formats and audio tracks.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -9112,5 +9112,10 @@
             "finish-args-host-filesystem-access": "SFTP client needs direct access to user-chosen local paths outside home (removable media, /mnt, /srv, project trees) for transfer, sync, and directory-watch operations. FileChooser portal is used where the user picks individual files, but background transfers and watched directories need persistent access to the chosen paths.",
             "finish-args-has-socket-ssh-auth": "Required to use the user's running SSH agent for public-key authentication against remote hosts (the standard mechanism for yubikey/gpg-agent/passphrase-protected keys)."
         }
+    },
+    "org.tahoma2d.Tahoma2D": {
+        "stable": {
+            "finish-args-host-filesystem-access": "Tahoma2D is regularly used in conjunction with other drawing or imagining applications and must be able to import and/or write image files whereever it needs to so it can be shared by other software involved in their animation pipeline."
+        }
     }
 }
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -9115,7 +9115,7 @@
     },
     "org.tahoma2d.Tahoma2D": {
         "stable": {
-            "finish-args-host-filesystem-access": "Tahoma2D is regularly used in conjunction with other drawing or imagining applications and must be able to import and/or write image files whereever it needs to so it can be shared by other software involved in their animation pipeline."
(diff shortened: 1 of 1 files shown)


CodedOre (changes requested):
> Tahoma2D is regularly used in conjunction with other drawing or imagining applications and must be able to import and/or write image files whereever it needs to so it can be shared by other software involved in their animation pipeline.

Sure. A lot of applications want to read/write files that also might be used by other applications. \
But why is it not possible to utilize the FileChooser portal to select a project folder to read/write from?

That's the point you need to really answer. Because if access over the portal would suffice, then this permission is not needed. 

> Additionally, some image and audio files require ffmpeg or rhubarb and users need to set the location of these executables in preferences for use by the application."

ffmpeg is already in the runtime. And if `rhubarb` is required, it should be bundled with the application. 

You should not rely on software that may or may not be installed on the host.

manongjohn:
My application currently doesn't support using XDG portals.

As a matter of convenience, I don't have any issues with bundling additional software with my release, though I have concerns about licensing so I tend to bundle with specific versions, even compiling them myself, in order keep to a certain license (like ffmpeg).  However due to that, the version I have bundled is sometimes not adequate for their needs (I may have compiled something out of it due to licensing concerns) and/or they need to use a different version that they install independently.  This tends to happen with ffmpeg.  We've seen cases where newer versions of ffmpeg were problematic and they were forced to use an older version or vice versa.  I do not want to restrict my users to a specific version, if a certain version suits their needs better for their animation.

Erick555:
Note that running external, user provided binaries like ffmpeg inside flatpak environment may not work even when you give access to the binary due to incompatible abi.

manongjohn:
> Note that running external, user provided binaries like ffmpeg inside flatpak environment may not work even when you give access to the binary due to incompatible abi.

I just confirmed locally that the flatpak build of my application was successfully able to use an external version of ffmpeg.

If it doesn't work for some people, then that's a limitation of their system they would have to work around. I don't want to hinder my users right off the bat from producing the best possible animations because I'm forcing them to use only what flatpak provides.

hfiguiere:
> I just confirmed locally that the flatpak build of my application was successfully able to use an external version of ffmpeg.

This is just a fluke. Thinking this is the actual expection show a great misunderstanding.

manongjohn:
> > I just confirmed locally that the flatpak build of my application was successfully able to use an external version of ffmpeg.
> 
> This is just a fluke. Thinking this is the actual expection show a great misunderstanding.

Not everyone is going to have your innate knowledge or level of how everything works. 

My apologies for being stupid, but users are going to try it if it's possible and it works, then it works.  If it doesn't, then it doesn't.  If you want to look down on me for it, so be it.  I could care less what you think about me. I care about my users and trying to do what I think is best for them.

Not sure what else to say or add to get this exception approved.

Btw, somehow this fluke works with flatpak version of OpenToonz also.

manongjohn:
My application cannot work with just the sandbox settings.  It cannot write and files to save or anything rendering my application completely useless.

If I reopen this and change the setting to `home`, will that be acceptable?

bbhtt:
> If I reopen this and change the setting to home, will that be acceptable?

No

manongjohn:
Why?

bbhtt:
I'm not required to give you detailed explanations, refer to the discussion above.

PR #1039, opened 2026-05-22
Add exception for com.devolutions.remotedesktopmanager

ChevalDentifrice:
App: https://github.com/flathub/com.devolutions.remotedesktopmanager
Upstream: https://devolutions.net/remote-desktop-manager/
Manifest PR: https://github.com/flathub/com.devolutions.remotedesktopmanager/pull/32

**finish-args-flatpak-spawn-access** is needed for two features in RDM:

1. For the Local Terminal. We have a builtin terminal feature. The shell is user configurable in the preferences. The point of this feature is to give the user their host shell with their PATH, ssh-agent and host tools. A sandboxed shell cant see these.

2. For PowerShell connection type in external mode: the pwsh session run inside the user's host terminal emulator (gnome-terminal, konsole, xterm, and more). The user picks which one in the preferences.. flatpak-spawn launches it, the emulator runs pwsh. It cant be sandboxed because the point of pwsh on Linux for our users is to access to host tools.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -569,7 +569,8 @@
     },
     "com.devolutions.remotedesktopmanager": {
         "stable": {
-            "finish-args-has-socket-ssh-auth": "Predates the linter rule"
+            "finish-args-has-socket-ssh-auth": "Predates the linter rule",
+            "finish-args-flatpak-spawn-access": "Required for Local Terminal and PowerShell to spawn host shells."
         }
     },
     "com.diffingo.fwbackups": {


bbhtt:
Can you provide some demo video to show the features? I don't have time to install the app to check it out.

What happens when the preferred terminal is also a Flatpak app (or snap)?

This is also a proprietary app and I would be more confident granting these exceptions to break away from the Flatpak sandbox if I could see the code

ChevalDentifrice:
Demo:


https://github.com/user-attachments/assets/540a3ec3-bd62-476a-b265-9c869c19b44f


At first it wouldnt have worked for terminal emulator installed via Flatpak, but I added support for it (you can see the test with Ptyxis in the demo).

About the code, here is the part where we use the flatpak-spawn:
TerminalHelper (2 wrappers):

```
public static string WrapShellCommandForFlatpak(string shellCommand)
{
    if (!ApplicationManager.IsFlatpak || string.IsNullOrEmpty(shellCommand))
    {
        return shellCommand;
    }

    string home = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
    return $"flatpak-spawn --host --directory=\"{home}\" {shellCommand}";
}

public static void WrapStartInfoForFlatpak(ProcessStartInfo startInfo)
{
    if (!ApplicationManager.IsFlatpak || startInfo == null)
    {
        return;
    }

    string home = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
    string originalFileName = startInfo.FileName;
    string originalArguments = startInfo.Arguments;

    startInfo.FileName = "flatpak-spawn";
    startInfo.Arguments = $"--host --directory=\"{home}\" \"{originalFileName}\" {originalArguments}";
}
```

Local Terminal:

`terminal.Shell = TerminalHelper.WrapShellCommandForFlatpak(OptionManager.Options.LocalTerminalShell);`

External Powershell:

```
ProcessStartInfo processStartInfo = TerminalHelper.CreateExternalStartInfo(powershellExecutable, powerShellArguments.ToString(), terminalEmulatorPath);
TerminalHelper.WrapStartInfoForFlatpak(processStartInfo);
Process process = new Process() { StartInfo = processStartInfo };
process.Start();
```

PR #1041, opened 2026-05-23
Add exception for io.github.voltkraft.immich-folder-watch

VoltKraft:
Avalonia's StatusNotifierItem tray integration owns a runtime-generated org.kde.StatusNotifierItem-<pid>-<id> bus name. The Flatpak D-Bus proxy only supports subtree wildcards, so org.kde.* is the narrowest pattern that matches the generated SNI name.


bbhtt:
This exception is no longer given. You have to disable the broken tray implementation.

Something that forces people to use a wildcard name to entirety of KDE services is a security issue and should never be allowed.

Qt had this issue too and they fixed it in Qt upstream long ago. It is a problem if certain frameworks haven't fixed themselves.

PR #1043, opened 2026-05-25
Add exception for org.koderoots.chatqt

DenysMb:
In ChatQT, we have some experimental features like using local MCPs, OpenCode and Pi inside the app. We need to use flatpak spawn for that.

Related PR: https://github.com/flathub/flathub/pull/8764

https://github.com/user-attachments/assets/67942cc3-ad90-48bd-9db9-9f078a10c317

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -7828,6 +7828,11 @@
             "finish-args-contains-both-x11-and-wayland": "Predates the linter rule"
         }
     },
+    "org.koderoots.chatqt": {
+        "stable": {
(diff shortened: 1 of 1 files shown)


CodedOre:
Instead of using `flatpak-spawn --host` to launch foreign binaries the user has to manually install, it would be better to provide the necessary binaries in the Flatpak environment. 

If you don't want to install everything with the application, you can use an extension point for the additional binaries. \
As a reference: Alpaca had done something similar for a while, providing Ollama in an separate extension. 

And if a pro-user would really want to use an binary outside the sandbox, they can add the override themselves. 

In short, I think there are better mechanisms than an exception to use `flatpak-spawn --host` for your use-case.

DenysMb:
Closed as the permission is not needed anymore.

PR #1046, opened 2026-05-25
exceptions: Update exceptions for WiVRn

ImSapphire:
flathub/io.github.wivrn.wivrn#40
https://github.com/flathub-infra/vorarbeiter/actions/runs/26386005464


hfiguiere:
I don't see why you would need this.

ImSapphire:
> I don't see why you would need this.

WiVRn supports body tracking via SlimeVR, for this we need to connect to SlimeVR's RPC socket to receive the body skeleton data.
When using the Steam version of SlimeVR, they create the sockets inside `$XDG_DATA_HOME/dev.slimevr.SlimeVR`. In the linked PR I provided a link to the code which handles that: https://github.com/SlimeVR/SlimeVR-Server/blob/61687a27ca1d5ee4eee98dc4b244b0687536d58e/server/core/src/main/java/io/eiren/util/OperatingSystem.kt#L42

PR #1047, opened 2026-05-28
Add DataFlow home filesystem access exception

sorguido:
This adds a lint exception for `io.github.sorguido.dataflow-procurement-software`.

DataFlow is a desktop procurement application that must preserve compatibility with its native Linux and Windows builds.

The application manages SQLite database persistence, Excel/PDF exports, database backups, attachments, templates, and user-selected files/directories. For the 2.3.0 Flatpak package, `--filesystem=home` is required to preserve the expected desktop-native behavior.

A stricter XDG-only setup was tested by removing `--filesystem=home` and granting only:

- `--filesystem=xdg-documents:create`
- `--filesystem=xdg-download:create`
- `--filesystem=xdg-desktop:create`

That setup removes the lint error, but it breaks the expected application behavior: DataFlow creates an alternative sandboxed database, and a newly created RFQ does not remain persistent after restarting the application.

A more sandbox-friendly variant would require explicit application-level path handling changes for database/config persistence and user-selected file handling, which is outside the scope of the current 2.3.0 Flatpak package.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -4746,6 +4746,11 @@
             "finish-args-host-filesystem-access": "needed to access and modify projects outside of sandbox"
         }
     },
+    "io.github.sorguido.dataflow-procurement-software": {
+        "stable": {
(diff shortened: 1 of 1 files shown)


CodedOre:
There is neither an released application nor an submission for an application with the id `io.github.sorguido.dataflow-procurement-software`.

So, this exception PR has no purpose at the moment. 

Also, to me it sounds you just don't want to be bothered to use the FileChooser portal in your application, which is a terrible reasoning for an exception to access all files in the home directory.

sorguido:
Thanks for the review. I’ll close this for now and revisit only if a Flathub submission is opened.

sorguido:
Thanks for the review. I’ll close this for now and revisit only if a Flathub submission is opened.

PR #1049, opened 2026-05-29
Add io.itch.nordup.TheGates x11+wayland exception

Nordup:
Requesting an exception to `finish-args-contains-both-x11-and-wayland` for **io.itch.nordup.TheGates** (already published on Flathub).

### Why both sockets are needed simultaneously

The app spawns two separate Godot processes:

- **Launcher UI** (trusted) — needs **X11/XWayland** because of native-Wayland UI bugs in the Godot version it ships.
- **Renderer** (loads untrusted remote "world" content) — deliberately run on **native Wayland** (`--display-driver wayland`) so that potentially malicious content cannot snoop or inject into the rest of the session via the shared X11 socket.

`--socket=fallback-x11` cannot satisfy this: on a Wayland session it hides the X11 socket entirely, but the launcher needs X11 **concurrently** with the renderer's Wayland. Hence both `--socket=x11` and `--socket=wayland` are required at the same time.

The exception is scoped to the `stable` repo. App manifest PR using both sockets: flathub/io.itch.nordup.TheGates#100.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -5069,6 +5069,11 @@
             "finish-args-own-name-wildcard-org.kde": "Still uses legacy StatusNotifier implementation"
         }
     },
+    "io.itch.nordup.TheGates": {
+        "stable": {
+            "finish-args-contains-both-x11-and-wayland": "The app spawns two separate Godot processes: a trusted launcher UI and a renderer process that loads untrusted remote world content. The launcher requires X11/XWayland because of native-Wayland UI bugs in Godot, while the renderer is deliberately run on native Wayland (--display-driver wayland) to isolate untrusted content from the rest of the session. fallback-x11 cannot satisfy this: on a Wayland session it hides the X11 socket, but the launcher needs X11 concurrently with the renderer's Wayland, so both sockets are required at the same time."
+        }
+    },
     "io.jamulus.Jamulus": {
         "*": {
             "finish-args-contains-both-x11-and-wayland": "Predates the linter rule"


bbhtt:
Godot itself defaults to X11 so the claim about security is dubious. You should only use `--socket=x11`.

Nordup:
Thanks! Fair question — this app is a bit unusual, so let me give the context.

TheGates is a "3D web browser": users open URLs that download third-party Godot worlds ("gates") and run them. Like a browser sandboxes tabs, we sandbox each world. It runs as two cooperating Godot processes:

- **Launcher** (the browser UI — menus, search, tabs): trusted, unsandboxed. It needs **X11/XWayland** because the Godot version it ships has native-Wayland UI bugs (popups/menus). Godot defaulting to X11 is exactly what the launcher wants.
- **Renderer** (runs the untrusted downloaded world): confined by a Firefox/Chromium-style sandbox — seccomp-bpf + Landlock + capability drop + `NO_NEW_PRIVS`, engaged before any gate code runs. It's explicitly launched with `--display-driver wayland` and pinned to **native Wayland** for Wayland's per-client isolation: a malicious world can't keylog or screen-capture the rest of the session the way an X11 client can. (The seccomp allowlist also omits `socket`/`connect`/`bind`, so the renderer can't open its own sockets — networking is brokered by the launcher.)

So both are needed at once: X11 for the trusted launcher, Wayland for the untrusted renderer. `fallback-x11` can't express this — on a Wayland session it hides the X11 socket, leaving the launcher with no XWayland.

Sandbox design and a runnable verification harness are public: https://docs.thegates.io/en/latest/about/security.html (and `tools/run-sandbox-test.py` in the engine fork). Happy to scope the exception to `stable` only.

bbhtt:
The exception was already scoped to public. This is clearly AI generated.

Nordup:
Yeah, I'm sorry

I wanted to say that for the renderer process, we launch it with --display-driver wayland
So it picks Wayland and cannot open any other sockets after lockdown (seccomp blocks all socket-related operations)

Tried Wayland only, but Godot support isn't there yet:(

Can you please reconsider?

bbhtt:
No, you should have thought twice before producing this wall of LLM text and making me read it.

Secondly, Godot does not default to Wayland because it still is not ready. You should not use `--display-driver wayland` in the first place to force it and then put yourself in a position where you need these meaningless permissions.

There is zero security improvement if you are mounting both x11 and wayland in the sandbox. 

This exception is only granted if the app has unsolvable bugs and causes crashes (like in proprietary apps etc.)

Nordup:
I'm really, really sorry, sir 🙏 I feel terrible and apologize
I messed up and won't bother you anymore

We've forked the Godot engine and have vendored the Chromium sandboxing model, so at a certain point the child process locks itself down before executing any untrusted code. So if godot opened only wayland socket before lockdown, it won't be able to open x11 even if it's mounted.

Fixed the LLM description in exceptions just in case:
https://github.com/Nordup/flatpak-builder-lint/commit/7052b9d48f061d0af91475476a112dae115fc6ed

But will take your no as an answer 🙏 sorry!

PR #1051, opened 2026-05-31
Add exceptions for org.atheme.audacious

ghost:
Required to put album covers into /tmp for MPRIS to pick them up.

Related PR: https://github.com/flathub/org.atheme.audacious/pull/62


bbhtt:
Any idea if it is possible to use a different directory in the app? Firefox uses something under .firefox. Extracring in tmp is not ideal for sandboxes.

You can also try setting TMPDIR to some XDG_* location in the sandbox

ghost:
You are right. There is another way. Thanks for the advice. Closed.

PR #1052, opened 2026-05-31
Add exception for com.axantum.xecrets.ez

xecrets:
In reference to https://github.com/flathub/flathub/pull/8813#pullrequestreview-4386903211.

~~Branch "*" (both stable and beta). If this is a problem, "stable" will have to do.~~

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -378,6 +378,11 @@
             "flathub-json-automerge-enabled": "Predates the linter rule"
         }
     },
+    "com.axantum.xecrets.ez": {
+        "*": {
+            "finish-args-host-filesystem-access": "Probes, creates, renames sibling files in bulk and non-interactively, i.e. 'file.txt <=> 'file-txt.axx', 'file (1).txt' etc."
+        }
+    },
     "com.axosoft.GitKraken": {
         "stable": {
             "finish-args-has-socket-ssh-auth": "Predates the linter rule",
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -379,7 +379,7 @@
         }
     },
     "com.axantum.xecrets.ez": {
-        "*": {
+        "stable": {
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -378,6 +378,11 @@
             "flathub-json-automerge-enabled": "Predates the linter rule"
         }
     },
+    "com.axantum.xecrets.ez": {
+        "*": {
```
CodedOre: You should use `stable` instead.

bbhtt:
the other thing i noticed is that should be home I think? and /media, /mnt /run/media if you need external drive access?

Rest of host is generally unwriteable and no one will be decrypting files there.

The reason this is not covered by a portal is there is no neighbouring files portal.

xecrets:
> Rest of host is generally unwriteable and no one will be decrypting files there.

Anywhere the user can see a writable file/folder is a possible location for encryption and decryption of files. If the user can see it and write it they should be able to encrypt or decrypt it. We shouldn't make assumptions about user behavior.

bbhtt:
Did you read what I said? Most of host is not writeable by the user because it is owned by root.

And people don't store random files needing decryption in /opt, /var etc. they are also not writeable either.

xecrets:
/tmp, /var/tmp, /mnt, /media, /run/media + home?

bbhtt:
What usecase involves decrypting files in /tmp or /var/tmp? /var/tmp is not covered by host either it is reserved by Flatpak and maps to ~/.var/app/$appid/tmp.

The other four should be enough I think.

xecrets:
> What usecase involves decrypting files in /tmp [omitted ...]

Many. Passing files between accounts. Often used as a place for experimental, temporary and ad-hoc work. Some prefer to use it when it is RAM based. ...and so on. It *is* writable. If used without static permission, risk of effective data loss. I am wary of restricting or predicting user behavior.

> /var/tmp is not covered by host either it is reserved by Flatpak [omitted ...]

Ok.

bbhtt:
Ok then update it to finish-args-host-tmp-access and finish-args-home-filesystem-access

PR #1053, opened 2026-06-02
Update permissions for com.bitwarden.desktop to enable shared unlock / browser integration

quexten:
https://github.com/bitwarden/clients/pull/14836 added support for the browser integration for flatpak, which enables the biometric unlock (and soon shared unlock) in the browsers, when the bitwarden desktop app is packaged as a flatpak. The PR for pulling in the upstream permissions into the flathub build https://github.com/flathub/com.bitwarden.desktop/pull/320 is currently failing since the linting rules require to add exceptions for these permissions.

This PR updates the exceptions, granting the flathub build access to the native messaging hosts directories of chrome, chromium, firefox.

We intentionally scoped these as narrowly as we can instead of giving the desktop app access to the entire home directory, and to poke as little of a hole into the flatpak sandbox as possible.

As to the "temporary solution" aspect to of the policy, once efforts such as the NativeMessagingPortal mature, we will switch to these, and can then remove these permissions. Currently, there are no standards-based / portal based alternatives.

---

I've also taken the liberty to edit the generic notes for predating the permission into giving an actual reason for the existing permissions.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -437,7 +437,14 @@
         "stable": {
             "appid-ends-with-lowercase-desktop": "app-id predates this linter rule",
             "finish-args-contains-both-x11-and-wayland": "X11 access is needed for clipboard access",
-            "finish-args-login1-system-talk-name": "System talk access is needed to enable biometric / polkit unlock"
+            "finish-args-login1-system-talk-name": "System talk access is needed to enable biometric / polkit unlock",
+            "finish-args-flatpak-appdata-folder-com.google.Chrome-config-google-chrome-NativeMessagingHosts-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-flatpak-appdata-folder-org.chromium.Chromium-config-chromium-NativeMessagingHosts-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-flatpak-appdata-folder-com.microsoft.Edge-config-microsoft-edge-NativeMessagingHosts-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-flatpak-appdata-folder-org.mozilla.firefox-.mozilla-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-unnecessary-xdg-config-microsoft-edge-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-unnecessary-xdg-config-chromium-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-unnecessary-xdg-config-google-chrome-rw-access": "Needed to enable shared unlock / biometric unlock browser integration."
         }
     },
     "com.bitwig.BitwigStudio": {
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -436,8 +436,8 @@
     "com.bitwarden.desktop": {
         "stable": {
             "appid-ends-with-lowercase-desktop": "app-id predates this linter rule",
-            "finish-args-contains-both-x11-and-wayland": "Predates the linter rule",
-            "finish-args-login1-system-talk-name": "Predates the linter rule"
+            "finish-args-contains-both-x11-and-wayland": "X11 access is needed for clipboard access",
+            "finish-args-login1-system-talk-name": "System talk access is needed to enable biometric / polkit unlock"
         }
     },
     "com.bitwig.BitwigStudio": {


bbhtt:
Any idea what's missing for NM to not need filesystem access for the manifests?

Afaik, Firefox and Chromium added support for XDG NM Proxy.

Is that something on their side, portal / proxy side or your side?

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -436,8 +436,15 @@
     "com.bitwarden.desktop": {
         "stable": {
             "appid-ends-with-lowercase-desktop": "app-id predates this linter rule",
-            "finish-args-contains-both-x11-and-wayland": "Predates the linter rule",
-            "finish-args-login1-system-talk-name": "Predates the linter rule"
+            "finish-args-contains-both-x11-and-wayland": "X11 access is needed for clipboard access",
+            "finish-args-login1-system-talk-name": "System talk access is needed to enable biometric / polkit unlock",
+            "finish-args-flatpak-appdata-folder-com.google.Chrome-config-google-chrome-NativeMessagingHosts-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
```
bbhtt, line 441: why are all of these and below RW?

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -436,8 +436,15 @@
...
+            "finish-args-flatpak-appdata-folder-org.chromium.Chromium-config-chromium-NativeMessagingHosts-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-flatpak-appdata-folder-com.microsoft.Edge-config-microsoft-edge-NativeMessagingHosts-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-flatpak-appdata-folder-org.mozilla.firefox-.mozilla-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-unnecessary-xdg-config-microsoft-edge-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
```
bbhtt, line 445: why is it subdir for Flatpak but the full data directories for host browsers?

flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -436,8 +436,15 @@
...
-            "finish-args-contains-both-x11-and-wayland": "Predates the linter rule",
-            "finish-args-login1-system-talk-name": "Predates the linter rule"
+            "finish-args-contains-both-x11-and-wayland": "X11 access is needed for clipboard access",
+            "finish-args-login1-system-talk-name": "System talk access is needed to enable biometric / polkit unlock",
+            "finish-args-flatpak-appdata-folder-com.google.Chrome-config-google-chrome-NativeMessagingHosts-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-flatpak-appdata-folder-org.chromium.Chromium-config-chromium-NativeMessagingHosts-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-flatpak-appdata-folder-com.microsoft.Edge-config-microsoft-edge-NativeMessagingHosts-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
+            "finish-args-flatpak-appdata-folder-org.mozilla.firefox-.mozilla-rw-access": "Needed to enable shared unlock / biometric unlock browser integration.",
```
bbhtt, line 444: why full .mozilla for Firefox?

quexten:
@bbhtt So there is three reasons currently for this solution:

1. Firefox does not ship with the xdg proxy enabled by default still:
- https://phabricator.services.mozilla.com/source/firefox-beta/browse/beta/modules/libpref/init/StaticPrefList.yaml$20223
2. We are looking for a stable implementation / user as reference. Currently it is rather hard to find good documentation on how to interact with the proxy.
3. It is unclear to me whether the portal approach supports Flatpak-to-flatpak communication, or whether it requires the host browser to be unsandboxed

The current solution at least unblocks the feature for most users, giving us room to research the portal solution / wait for it to be more mature or documented.

flatpak_builder_lint/staticfiles/exceptions.json
quexten: Firefox by default doesn't create the NativeMessagingHost directory. In the flathub manifest, we have the following note:
```
  # Firefox does not create NMHS directory by default
  - --filesystem=~/.var/app/org.mozilla.firefox/.mozilla/
```

flatpak_builder_lint/staticfiles/exceptions.json
quexten: We need to place a few files in the native messaging directory, namely:
- The native messaging manifest
- A unix socket
- A proxy executable

The browser then (in the sandbox if it is flatpak) looks at the native messaging manifest, runs the executable and the executable connects to the unix socket which allows it to talk to the desktop app. To create the socket, and the executable, they need to be `rw`.

This is different from how the app used to behave, since we could previously have the executable and unix socket live in the app owned directory. However, other browsers generally are also sandboxed and cannot see executables that the bitwarden app owns.

bbhtt:
Ok thanks.

flatpak_builder_lint/staticfiles/exceptions.json
quexten: I just took the errors that vorarbeiter lint job gave us. The permissions are narrowly scoped to the NMHS dirs. From the manifest yaml:
```
  # Non-sandboxed
  - --filesystem=xdg-config/google-chrome/NativeMessagingHosts/
  - --filesystem=xdg-config/chromium/NativeMessagingHosts/
  - --filesystem=xdg-config/microsoft-edge/NativeMessagingHosts/
  - --filesystem=home/.mozilla
  # Flatpak-sandboxed
  - --filesystem=~/.var/app/org.chromium.Chromium/config/chromium/NativeMessagingHosts/
  - --filesystem=~/.var/app/com.google.Chrome/config/google-chrome/NativeMessagingHosts/
  - --filesystem=~/.var/app/com.microsoft.Edge/config/microsoft-edge/NativeMessagingHosts/
  # Firefox does not create NMHS directory by default
  - --filesystem=~/.var/app/org.mozilla.firefox/.mozilla/
```

quexten:
Thank you for the quick review and merge!

bbhtt:
Sandboxed NM host or sandboxed NM host to sandboxed browser seems like an unsolved problem https://github.com/flatpak/xdg-native-messaging-proxy/issues/9

(forgot to mention this).

PR #1056, opened 2026-06-07
Add finish-args-has-dev-usb exception for org.gnome.gedit

Pointedstick:
See https://github.com/flathub/org.gnome.gedit/pull/172.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -6781,7 +6781,8 @@
     "org.gnome.gedit": {
         "stable": {
             "appstream-id-mismatch-flatpak-id": "appstream-cid predates this linter rule",
-            "finish-args-host-filesystem-access": "Predates the linter rule"
+            "finish-args-host-filesystem-access": "Predates the linter rule",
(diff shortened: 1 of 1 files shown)


hfiguiere:
1. I don't see how device=usb would help.
2. This is only available on very recent flatpak so it's not accepted on flathub as most distributions in use don't have it. device=all should be used instead. But see 1.

Pointedstick:
> I don't see how device=usb would help.

What do you mean exactly? When I add this permission, USB-connected removable devices become visible in the sidebar of the non-portal open/save dialogs. Without it, they don't.

I can change it to `device=all` if that's preferred.

bbhtt:
It's behind an exception because it's not backwards compatible.  

You should use `all` for now.

Flatpak 1.18.0 coming soon will allow using it in a backwards compatible manner using the new conditional permission syntax (see Flatpak docs).

See also https://github.com/flathub-infra/flatpak-builder-lint/pull/1015

bbhtt:
Fwiw I am also not sure why removable USB devices require device access. You would only need filesystem access to their mountpoints to browse the files which is /run/media, /media and /mnt but gedit seems to have `host` already.

PR #1058, opened 2026-06-13
Add X11+Wayland exception for com.dekomote.vermouth

dekomote:
A lot of older Linux games come with old SDL bundled which doesn't have a wayland driver and they fail to launch if the X11 socket is unavailable on a wayland system. Furthermore, Proton only has experimental wayland support and it's often buggy.


bbhtt:
This doesn't satisfy the criteria https://docs.flathub.org/docs/for-app-authors/linter#finish-args-contains-both-x11-and-wayland

You should use only socket=x11 until everything is ready for wayland.

PR #1063, opened 2026-06-18
Add finish-args-contains-both-x11-and-wayland exception for com.orcaslicer.OrcaSlicer

SoftFever:
OrcaSlicer now supports Wayland. 
However, the GUI currently freezes and becomes unresponsive when using `--socket=wayland` and `--socket=fallback-x11` simultaneously.

Tested with: https://github.com/flathub/com.orcaslicer.OrcaSlicer/pull/3

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -9051,7 +9051,9 @@
     "com.orcaslicer.OrcaSlicer": {
         "stable": {
             "finish-args-home-filesystem-access": "OrcaSlicer needs persistent access to model files, recent projects, and G-code export across arbitrary user directories",
-            "finish-args-flatpak-appdata-folder-io.github.softfever.OrcaSlicer-ro-access": "Read-only access to legacy config/cache directory (~/.var/app/io.github.softfever.OrcaSlicer) for migration from old app-id"
+            "finish-args-flatpak-appdata-folder-io.github.softfever.OrcaSlicer-ro-access": "Read-only access to legacy config/cache directory (~/.var/app/io.github.softfever.OrcaSlicer) for migration from old app-id",
+            "finish-args-contains-both-x11-and-wayland": "The app supports Wayland, but currently, the GUI freezes and becomes unresponsive when using --socket=wayland and --socket=fallback-x11 simultaneously."
+
         }
     },
     "org.cvfosammmm.Lemma": {


CodedOre:
> However, the GUI currently freezes and becomes unresponsive when using --socket=wayland and --socket=fallback-x11 simultaneously.

That sounds to me like the Wayland implementation has an issue. \
Such an issue should rather be resolved than being swept under the rug. 

Also, if `--socket=x11` makes it work, are you sure your application is not just using X11 instead of Wayland and has not the issues because of this?

So, for the time being, its probably best to stay with X11 only until the Wayland implementation works properly.

SoftFever:
> > However, the GUI currently freezes and becomes unresponsive when using --socket=wayland and --socket=fallback-x11 simultaneously.
> 
> That sounds to me like the Wayland implementation has an issue. Such an issue should rather be resolved than being swept under the rug.

I agree that it's likely an issue with the Wayland implementation. I can't determine whether the problem is caused by our code or one of the third-party UI components we use, as we rely on quite a few of them. I'm running out of ideas, so improving Wayland support remains an ongoing long-term effort.

> 
> Also, if `--socket=x11` makes it work, are you sure your application is not just using X11 instead of Wayland and has not the issues because of this?

I can confirm it is running on Wayland; I used xlsclients and xeyes to verify this.

> 
> So, for the time being, its probably best to stay with X11 only until the Wayland implementation works properly.

Here is the dilemma: I added Wayland support to address some issues that only occur on X11 (e.g., multiple monitor support), and it works fine on Wayland.  
So far, the community has tested extensively with Wayland (using `--socket=wayland` and `--socket=x11`), and while it's not perfect, it is reasonably stable and generally better than X11.

bbhtt:
This doesn't fit the criteria for granting this exception. You should either use x11 only ir fix the bug in your wayland implementation.

We aren't granting exceptions for unfixed bugs. That's bypassing the problem.

SoftFever:
> This doesn't fit the criteria for granting this exception. You should either use x11 only ir fix the bug in your wayland implementation.
> 
> We aren't granting exceptions for unfixed bugs. That's bypassing the problem.

Can you reconsider? I don’t think this should be classified as an unfixed bug. Otherwise, by that criterion, the exception shouldn’t exist in the first place—by that logic, no software should ever request it due to the crashes or unsupported features.

bbhtt:
According to your own words:

> I agree that it's likely an issue with the Wayland implementation. 

So the correct thing to do here is to investigate and/or fix the bug in your app, not ask for an exception. So no.

bbhtt:
fwiw I can reconsider if the app is unusable with both `--socket=x11` and fallback-x11+wayland. Unusable means crashing freezing etc. and you also have to show that the bug is in some component other than your app and file a proper issue for it. That's the documented criteria.

Anything other than that would be bypassing the actual problem and juggling between different sets of bugs between X11 and Wayland. The exception is band-aid here as the original issue will not be fixed if it is not investigated properly.

PR #1064, opened 2026-06-18
Add finish-args-flatpak-spawn-access exception for io.kinvolk.Headlamp

yolossn:
Headlamp is a Kubernetes desktop UI. Cluster authentication relies on kubeconfig exec credential plugins (client-go), which invoke host binaries such as `kubelogin` (Azure AD), the `aws` CLI (EKS), `gke-gcloud-auth-plugin` etc. These binaries are not present inside the sandbox, so `flatpak-spawn` access to the host is required for auth to work with different providers.

Corresponding manifest PR: flathub/io.kinvolk.Headlamp#77

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -5058,7 +5058,8 @@
     },
     "io.kinvolk.Headlamp": {
         "stable": {
-            "finish-args-contains-both-x11-and-wayland": "Predates the linter rule"
+            "finish-args-contains-both-x11-and-wayland": "Predates the linter rule",
+            "finish-args-flatpak-spawn-access": "Required to spawn kubelogin/az/aws/helm etc for headlamp to work with different auth providers"
         }
     },
     "io.kopia.KopiaUI": {


CodedOre:
What speaks against bundling the tools in the Headlamp Flatpak? Or in extensions, as proposed in https://github.com/flathub/io.kinvolk.Headlamp/pull/70?

It would be way better to make it work inside the sandbox, rather than to work around it by blowing a giant hole into the sandbox.

bbhtt:
Consider bundling in the app or as extensions.

yolossn:
Bundling makes Headlamp heavier, and we're trying to keep it lean. Most users already have these tools installed anyway, so wrapping something they already have into an extension feels like extra maintenance for us and wasted space for them, also we'd have to make a release every time one of those tools updates.

Extensions fix the size part, but they hit a different wall: the list of tools isn't fixed, it keeps growing. A kubeconfig can call basically any command for auth via exec plugins, tsh, aws-iam-authenticator, kubelogin, even some custom binary a company wrote internally. We can't ship an extension for a tool we've never heard of and it's not always just about the binary being there, plenty of these involve interactive logins (az login) or tools that call other tools (kubelogin shelling out to az), so an in-sandbox copy often can't handle the complete auth anyway.

CodedOre:
I still think using an extension point would be better. 

Even if the extensions you provide would only cover the majority of possible cases, that already means a lot of cases where a sandbox hole is not needed. 

Also, a user could theoretically build a custom extension locally to add company-specific or niche tools. 

If it would be possible to use extensions for most cases, limiting the use of flatpak-spawn to a few exceptions, then you would not need the permissions by default, but can let users do it as an override for those cases that really require it.

yolossn:
Just got to know that there is an effort to handle this the extenions way by an headlamp user. Can we get some eyes on that please?
https://github.com/flathub/flathub/pull/8611

bbhtt:
It's blocked on merging https://github.com/flathub/io.kinvolk.Headlamp/pull/70 first, so not our eyes needed there first.

yolossn:
From the discussion in https://github.com/flathub/flathub/pull/8611, it's not clear whether you're on board with the extension approach, it looks like @dylanmtaylor is still waiting to hear back from you. If we're aligned on going that route, I can start by getting https://github.com/flathub/io.kinvolk.Headlamp/pull/70 merged first.

bbhtt:
Yes, I am fine with the extension. I was just asking to make sure it makes sense to have one which was clarified later.

PR #1068, opened 2026-06-23
Add exception for org.gabmus.hydrapaper

ghost:
This linter error prevents me from updating org.gabmus.hydrapaper to a supported version.

The last release of this application was three years ago, and this application predates the linter rule.

https://github.com/flathub/org.gabmus.hydrapaper/pull/17


bbhtt:
Are you sure this is needed? These were supposed to be migrated https://github.com/flathub/flathub/issues/1040 I see it has the metadata for that but also this permission which seems confusing.

Also this seems like a case where the app should be EOLed? Or do you want to maintain it?

ghost:
Thanks, I removed related config paths. I can do basic maintenance until it's completely broken.

PR #1069, opened 2026-06-24
Add exception for app.legcord.Legcord

imide:
Running through native Wayland instead of X11 causes issues for a considerable number of users, but Wayland access is still needed for pipewire screen capturing

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -38,6 +38,11 @@
             "finish-args-unnecessary-xdg-data-typst-ro-access": "Well known common location of the user's Typst packages in the @local namespace"
         }
     },
+    "app.legcord.Legcord": {
+        "stable": {
+            "finish-args-contains-both-x11-and-wayland": "Running through native Wayland instead of X11 causes issues for a considerable number of users, but Wayland access is still needed for pipewire screen capturing"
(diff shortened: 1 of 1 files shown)


hfiguiere:
either you fix the crashes or you disable wayland.

imide:
> either you fix the crashes or you disable wayland.

the flatpak is based off of Vesktop, which itself is based off Armcord, and they all have this issue. Vesktop got an exception. in the README of the flatpak manifest, it explains how to enable wayland. (https://github.com/imide/flathub/tree/legcord-submission)
edit: armcord is the previous name of legcord prior to the CnD

bbhtt:
> Running through native Wayland instead of X11 causes issues

What issues? This is only allowed if it is unusable or causes crashes and you have to investigate / fix the issue if it is in your app or point us to an open issue if it is in some secondary component.

imide:
> > Running through native Wayland instead of X11 causes issues
> 
> What issues? This is only allowed if it is unusable or causes crashes and you have to investigate / fix the issue if it is in your app or point us to an open issue if it is in some secondary component.

from the looks of it, it seems all of the issues that required the workaround which also needed this exception have been fixed. i was operating under the "if it aint broke, dont fix it" mentality 😅 i'll be closing this PR and undoing the workaround. thanks for pointing this out to me!

PR #1073, opened 2026-07-02
Add exceptions for org.waywallen.waywallen

hypengw:
Required for loading image/video wallpapers from flatpak steam.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -8241,6 +8241,11 @@
             "finish-args-home-filesystem-access": "Predates the linter rule"
         }
     },
+    "org.waywallen.waywallen": {
+        "stable": {
+            "finish-args-flatpak-appdata-folder-com.valvesoftware.Steam-data-Steam-rw-access": "Required to allow user to use wallpapers downloaded in steam workshop"
+        }
+    },
     "org.wezfurlong.wezterm": {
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -8241,6 +8241,11 @@
             "finish-args-home-filesystem-access": "Predates the linter rule"
         }
     },
+    "org.waywallen.waywallen": {
+        "stable": {
+            "finish-args-flatpak-appdata-folder-com.valvesoftware.Steam-data-Steam-rw-access": "Predates the linter rule",
```
CodedOre: I don't see an application that is still in the submission process predating the linter rule...

Also: Why write access to the Steam folder?
If you only want to load wallpapers, why do you need to write to the Steam folder?

hypengw: https://github.com/flathub/flathub/pull/8804.

>Why write access to the Steam folder?

It allows user to use wallpapers downloaded in steam workshop.

CodedOre: Still, the linter rule is there for some time. Your submission is not old enough that it could claim the exception "predates the linter rule". You need an actual reason.

hypengw: Reason added.

PR #1078, opened 2026-07-18
Add finish-args-flatpak-spawn-access exception for io.github.zoutmax.FifineControlDeck

ZoutMax:
For the new-app submission flathub/flathub#9390 (fifine Control Deck — a native Linux controller for fifine AmpliGame D6 / Mirabox Stream Dock macro keypads).

The app's core purpose is executing user-configured key actions on the host: launching applications, running commands, sending hotkeys via host ydotool/xdotool, and media/volume via playerctl/wpctl — all via `flatpak-spawn --host`. Without it the app can only render key images.

Precedent: com.core447.StreamController (same app category, "required to allow plugins to open apps on the host"), app.devsuite.Ptyxis, ca.vtrlx.Telepipe, and other host-action apps carrying this exception.

The submission's test build currently fails only on this linter check.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -4694,6 +4694,11 @@
             "finish-args-home-filesystem-access": "Predates the linter rule"
         }
     },
+    "io.github.zoutmax.FifineControlDeck": {
+        "stable": {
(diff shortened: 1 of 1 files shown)


ZoutMax:
Withdrawing for now — the corresponding Flathub submission (flathub/flathub#9390) is parked until the project has more development history, so this exception request is moot. Will re-file if/when the submission returns.

PR #1080, opened 2026-07-18
Add X11 exception for org.catacombing.charon

chrisduerr:


chrisduerr:
See https://github.com/flathub/flathub/pull/9385.

PR #1081, opened 2026-07-18
Add mimeapps.list:ro exception for com.fightcade.Fightcade

Pobega:
Fightcade's Flatpak needs read-only access to ~/.local/share/applications to detect whether a conflicting non-Flatpak ("desktop") Fightcade install is also present. When both are installed they contend over the fightcade:// URL-scheme handler and challenge URLs stop launching correctly (see flathub/com.fightcade.Fightcade#126).

The access is strictly read-only and used only to check for two specific files (Fightcade.desktop, fcade-quark.desktop) and show a zenity warning. Nothing is written or executed.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -617,7 +617,8 @@
     },
     "com.fightcade.Fightcade": {
         "*": {
-            "flathub-json-automerge-enabled": "Predates the linter rule, outdated clients cannot function"
+            "flathub-json-automerge-enabled": "Predates the linter rule, outdated clients cannot function",
+            "finish-args-unnecessary-xdg-config-mimeapps.list-ro-access": "Read-only access to the host's mimeapps.list is used solely to detect a conflicting non-Flatpak Fightcade install at startup, by checking whether the fcade:// URL-scheme default handler is one other than the Flatpak's own com.fightcade.Fightcade.fcade-quark.desktop, then warn the user, since running both installs breaks fcade:// challenge-URL launching (flathub/com.fightcade.Fightcade#126). No files are written or executed."
         }
     },
     "com.flashforge.FlashPrint": {


CodedOre (commented):
I don't think that this is sufficient for an exception. 

Yes, multiple versions of the same application can interfere with each other. \
But I assume that this should, for a normal user, be an very niche situation to be in, that it not warrants an exception for all users.

Pobega:
I wouldn't consider it a niche situation. The non-Flatpak version from https://fightcade.com isn't a normal package - it's a downloadable tar.gz that upon first execution creates these .desktop files.

Unfortunately those completely supercede the Flatpak (they load first as URI handlers) and cause our package to completely break. As the primary support giver for the Linux version on the Fightcade Discord it's the #1 problem users run into. It happens often enough that it has its own section of the README on GitHub.

Unfortunately a lot of users either don't think to check the GitHub or don't join the Discord for support, and when the issue happens it's silently broken (and users may just believe the Flatpak itself is broken.)

Another side effect is that because the upstream package is just a .tar.gz it has no uninstall method; so unless the user knows to clear our their .desktop files they will stay after removing Fightcade (and if they only remove Fightcade.desktop and leave fcade-quark.desktop we remain broken.)

The issue is relevant enough that I reported the bug 3 years ago, and still came around to implementing it in 2026.

Erick555:
This permission isn't useful for flatpak itself and I think cleaning out the mess on user system may be not in scope of flatpak apps.

Pobega:
That's a fair position, but just to be clear we aren't cleaning the mess on user's system; this is purely to add a notification to alert a user that their system is misconfigured (and in the cases it fires, the application will **always** be broken.)

As the primary maintainer and support giver this is more to make my life easier (and inform users who are not on the Discord/reading the README.md), but if Flatpak wants to deny this permission I'll (begrudgingly) accept that.

I personally believe this is a pretty low stake and scoped change with minimal impact on user privacy.

Pobega:
Would using `--filesystem=xdg-config/mimeapps.list:ro` be preferable to `applications`? I can theoretically use that to differentiate between `com.fightcade.Fightcade.fcade-quark.desktop` and `fcade-quark.desktop`

similar to `xdg-mime`;

```
pobega:~ xdg-mime query default x-scheme-handler/fcade
com.fightcade.Fightcade.fcade-quark.desktop
```

barthalion:
mimeapps.list sounds acceptable to me.

Pobega:
Updated the PR here to use `mimeapps.list:ro`, and updated the PR in the other repository as well. Tested that it works against a default Fightcade install.

PR #1088, opened 2026-07-22
feat: add exception so bot can automerge

matheus1lva:
Add exception for automerge enabled on stable and beta channels for https://github.com/flathub/io.tether.QVAC.Workbench

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -8750,10 +8750,10 @@
     },
     "io.tether.QVAC.Workbench": {
         "stable": {
-            "flathub-json-automerge-enabled": "Predates the linter rule"
+          "flathub-json-automerge-enabled": "Convenience and easy to automate the whole deployment process"
         },
         "beta": {
-            "flathub-json-automerge-enabled": "Predates the linter rule"
+          "flathub-json-automerge-enabled": "Convenience and easy to automate the whole deployment process"
         }
 
     }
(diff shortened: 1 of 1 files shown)


CodedOre (changes requested):
This change does not "predate the linter rule", so there needs to be an actual reason here.

Why is it necessary to automerge the update PRs?

matheus1lva:
Ah i didnt find anyone else adding that so i went with what generate_exception gave me.

@CodedOre pure convenience. The automated version pulling is added already - now merging the pr automatic is a good to have.

CodedOre:
I personally can't say the convenience for you is proper reason enough to warrant that exception to be given.

matheus1lva:
> I personally can't say the convenience for you is proper reason enough to warrant that exception to be given.

@CodedOre Well, what other reasonable reasons would fall into the exception here? If thats available to use? The auto merge and auto pr creation is already a convenience just by these features existing.

CodedOre:
I'll let our infrastructure person do the ultimate call when he gets the time for it.

But automerging is to my understanding mainly for cases where it's vital to quickly roll out updates. Like for example Telegram which requires clients to be on the newest version.

It also comes with it's own issues, like no chance to do QA on the build (which you're supposed to do) and, if configured badly, putting a huge strain on the infrastructure.

So, making the exception just for convenience doesn't sound like an compelling reason.

barthalion:
Automerging is limited to extra-data apps as they tend to be legitimately broken due to older data being removed. Convenience isn't an excuse for anyone else; you are supposed to test your app before merging the PR.

PR #1090, opened 2026-07-24
Add exception for io.github.cosmic_utils.camera

FreddyFunk:
The upcoming version v1.0.0 of the Camera application added a feature to prevent the system from sleeping and locking when the Camera application is in use.


CodedOre (commented):
If this is about inhibiting screen lock and suspend, why not use the inhibit portal instead?

FreddyFunk:
For desktop environments that do not have the inhibit portal properly implemented. But maybe that's nothing Flathub should have to worry about. So I'm closing this now and update my Camera app to expect a proper inhibit portal implementation to get this feature working.

PR #1091, opened 2026-07-24
Filesystem exception for org.mpz_player.mpz

olegantonyan:
Without full filesystem access this player becomes too crippled to justify its existence on Flathub:
- music can be scattered all over the place, on network shares, external drives, etc, I have 4 different folders, for example
- QFileSystemWatcher uses inotify, which is needed to pick up the changes in library folders

It's been around for many years with existing userbase
https://github.com/olegantonyan/mpz

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -7683,7 +7683,7 @@
     },
     "org.mpz_player.mpz": {
         "stable": {
-            "finish-args-host-filesystem-access": "Folder-based music player: users add arbitrary directories (external drives, NFS-shares, etc) which mpz reads, watches for changes (QFileSystemModel, inotify), plus there is built-in tag editor and playlists export"
+            "finish-args-home-filesystem-access": "Folder-based music player: reads and watches user-picked music dirs recursively (QFileSystemModel, inotify), which portals cannot do; also tag editing and playlist export"
         }
     },
     "org.musescore.MuseScore": {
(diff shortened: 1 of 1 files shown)


CodedOre (commented):
Why do you require access to the full host though?

Would it work to just use `home` and external drives instead?

olegantonyan:
> Why do you require access to the full host though?
> 
> Would it work to just use `home` and external drives instead?

This has few issues:
1. reliably detect external drives. Is it `/mnt`? `/media`? `/run/media/$USER`? I've seen the transition between all these being the defaults on LInux, plus I used to have NFS shares mounted at `/storage`, and it's not uncommon for people to mount stuff at obscure locations.
2. having rw ~ and handful of /mnt, /media is already like 99% of sensitive places on most systems. If my app gets hacked it's very likely that the hacker will be interested in crypto wallets, browser's cache, stuff like that in ~, not /etc/shadow or something
3. filesystem changes monitoring that works on inotify, but it has issues with dekstop portal https://github.com/flatpak/xdg-desktop-portal/issues/567

CodedOre:
I will have to ask you to use `home`, `/mnt`, `/media` and `/run/media` instead. 

Filesystem permissions should be as strict as possible. So if those work for 99% of cases, that would be better.

olegantonyan:
It does not fix `inotify` issues. Filesystem access is the core feature of this app

Erick555:
How access to home, /mnt, /media and /run/media doesn't fix inotify?

olegantonyan:
Ok, correct, changes in these folders are inotified.
Folders outside of home, /mnt, /media and /run/media can still be used when selected via file selector (the only way in UI to add music library), but do not get inotify.

May I ask why it has to be so strict? Some distros (https://repology.org/project/mpz/versions) have no issues getting this app voluntarily into repos without filesystem restrictions, but flatpak version has to be crippled. Why?

barthalion:
"Some distros" run random software completely unsandboxed. Of course they don't care.

olegantonyan:
Updated exception to `finish-args-home-filesystem-access` b/c now it trips over this

PR #1097, opened 2026-08-02
Add exceptions for io.github.Faugus.faugus-launcher

Faugus:


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -3307,7 +3307,8 @@
             "finish-args-contains-both-x11-and-wayland": "Without socket=X11 Winetricks doesn't work on Wayland and without socket=wayland the Wine Wayland driver doesn't work.",
             "finish-args-home-filesystem-access": "Predates the linter rule",
             "finish-args-unnecessary-xdg-config-MangoHud-ro-access": "Required by MangoHud to detect the settings file.",
-            "finish-args-own-name-wildcard-com.steampowered": "Required by umu-launcher to access the Steam Runtime's pressure-vessel."
+            "finish-args-own-name-wildcard-com.steampowered": "Required by umu-launcher to access the Steam Runtime's pressure-vessel.",
+            "finish-args-unnecessary-xdg-data-themes-ro-access": "Required for detecting themes installed in ./local/share/themes."
```
CodedOre, line 3311: What kind of themes are you trying to access in `xdg-data/themes`?

If you want to access GTK3 themes, there is an extension point expected to be used, not an permission.

Faugus: GTK4 themes

CodedOre: As far as I can tell you use libadwaita as your platform library, which has no theming support in an official matter.

So, where is the theming done by?

If this is an feature by the Faugus launcher, the themes should be either an Faugus extension or in the Faugus configuration folder. 

If its an general theme modification done by some users, then I don't see why this would warrant an permission for all users. Permissions overrides exists.

Faugus: It can use Adwaita and GTK4 themes installed in the system.
The native package reads `/usr/share/themes`, `~/.themes` and `~/.local/share/themes`.

PR #1101, opened 2026-08-08
Add exceptions for io.github.Faugus.faugus-launcher

Faugus:
This is for the optional system tray icon. When the main window is closed with tray enabled, the app hands off to a tiny helper process instead of staying fully resident just to show an icon.

The only command that ever runs through `flatpak-spawn` is:
`flatpak-spawn --host flatpak run --command=python3 io.github.Faugus.faugus-launcher -m faugus.tray_only`

Always this exact string, always this app's own id, nothing user-controlled goes into it. It's not arbitrary host access, just relaunching through flatpak under the same sandbox restrictions as a normal launch.

The separate instance is needed because a plain child process (no special permission, sharing the main app's sandbox) doesn't work: closing the window causes the systray to draw a second, dead icon alongside the real one. A few things were tried before reaching for `flatpak-spawn`: Reordering the D-Bus environment, making it idempotent, a graceful quit instead of a hard exit, requesting the Background portal. None of that got rid of it; only giving the helper its own sandbox instance did.


CodedOre:
That is a hard no.

You want the ultimate sandbox escape… to fix an bug in your application. 

There should be a solution that doesn't require the sandbox escape. Like always running the tray icon from the small binary and launching the small binary with the UI.

PR #1106, opened 2026-08-10
Add exception for me.chocolateimage-graphics-creator

chocolateimage:
Project files contain paths to assets that are located anywhere on disk (can be images, and other assets too).

The project file can also be modified to have any path as an asset file (for example if you make a script that automatically changes the file and then render automatically).

This is why this exception is needed.


chocolateimage:
Seems like a Discord message is enough for my PR to get closed.

PR #1107, opened 2026-08-10
Add finish-args-host-var-access exception for io.github.idescriptor.iDescriptor

uncor3:
Hi, this exception is required for https://github.com/flathub/flathub/pull/9691

**Why**
- usbmuxd doesn't handle/recognize wireless devices on Linux and Windows.
- usbmuxd doesn't provide pairing record enumeration directly (meaning there is no way I can see what pairing files the user has through usbmuxd)
- the app must be able to read the pairing files so we can connect

**Solution**: `--filesystem=/var/lib/lockdown:ro` (readonly)


uncor3:
@CodedOre can you please review?

teamcons:
alright

PR #1112, opened 2026-08-12
Add finish-args-flatpak-appdata-folder-org.chnapy.pkvault-ro-access exception for io.github.chnapy.pkvault

Chnapy:
This app is currently used as a flatpak file with an ID `org.chnapy.pkvault`.
To follow Flathub rules, I changed this ID to `io.github.chnapy.pkvault`, but to avoid user data loss I need to migrate data from old ID to new one.

Related Flathub PR: https://github.com/flathub/flathub/pull/9743

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -3583,6 +3583,11 @@
             "finish-args-host-filesystem-access": "Predates the linter rule"
         }
     },
+    "io.github.chnapy.pkvault": {
+        "stable": {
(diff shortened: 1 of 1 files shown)


teamcons:
To clarify, the old ID was distributed as a bundle to Chnapy's users. This would be to ensure a smooth transition from the "off-flathub" bundle to the flathub one

I am fine with this!

teamcons:
By the power bestowed upon me i grant this

Please once enough time has passed according to you, remove the permission from your manifest.
Dont worry about the linter, exceptions gets purged when detected as unused for awhile

PR #1113, opened 2026-08-13
Add flathub-json-automerge-enabled exception for cn.eeo.ClassIn

imngkhang:
Hi Flathub reviewers, this exception is required for https://github.com/flathub/flathub/pull/9708 

**Why**
- This is a 3rd-party wrapper, and Classin is proprietary. 
- Because that, I used extra-data for ClassIn, and EEO use a repo that stays on their web, and I want to enable automatically flathubbot prs for the app can automatically update.

So can you review again and add this expection?

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -277,7 +277,8 @@
     },
     "cn.eeo.ClassIn": {
         "stable": {
-            "finish-args-home-filesystem-access": "Needed because the proprietary Qt-based file picker cannot access through symlinks, and allow users to select files from the root of their home directory."
+            "finish-args-home-filesystem-access": "Needed because the proprietary Qt-based file picker cannot access through symlinks, and allow users to select files from the root of their home directory.",
+            "flathub-json-automerge-enabled":"Allow merging of autobot PRs to keep ClassIn updated (uses extra-data with external link)"
         }
     },
     "cn.ottercorp.xivlaunchercn": {
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -278,7 +278,7 @@
     "cn.eeo.ClassIn": {
         "stable": {
             "finish-args-home-filesystem-access": "Needed because the proprietary Qt-based file picker cannot access through symlinks, and allow users to select files from the root of their home directory.",
-            "flathub-json-automerge-enabled":"Allow merging of autobot PRs to keep ClassIn updated (uses extra-data with external link)"
+            "flathub-json-automerge-enabled": "Allow merging of autobot PRs to keep ClassIn updated (uses extra-data with external link)"
         }
     },
     "cn.ottercorp.xivlaunchercn": {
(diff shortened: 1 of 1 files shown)


imngkhang:
it is over 48 hours, can you recheck this PR?

imngkhang:
Hi, can you take some times to review this expection? @teamcons

teamcons:
why do you need automerge though? You are supposed to test new builds before merging

imngkhang:
because I don't have time to maintain it (I only have time to maintain it on summer vacation), and now I don't have any co-maintainer to let them maintain with me, so I need it to reduce time (I also will test if the upstream release new version too and I have free time, because the Wayland problem)

imngkhang:
I think they are freezing or go to a major release, and a minor release never break (they use the same core), so it will only break if the upstream remove x11 support (which will be never in the next release). But runtime bumping I will do that manually, because I don't want to break the app by the runtime only because auto bumping

teamcons:
When submitting an app to flathub you take the responsibility to maintain it

We cant just let stuff be added, abandoned, and autobumped hoping for the best

imngkhang:
I will maintain it, that previous response I meant the stability of minor version, it will not break if it autobumped, and I will bump the runtime manually. If it over 4 months without a new version I will turn off automerge, and I will try to find a co-maintainer as my best to maintain it, if you don't agree

teamcons:
Wouldnt it make more sense, though, to ignore minor versions, so the flathub is slightly behind official releases, but is at least tested and not mystery package?

ive had stuff break with minor version bumps. I am not super comfortable with untested stuff going around. It is fine if not perfectly up to date

imngkhang:
> The automerge-flathubbot-prs feature has been introduced back in 2019 to reduce maintenance overhead of applications dependent on non-redistributable data. If [an extra-data source](https://docs.flatpak.org/en/latest/module-sources.html#extra-data) used by the app disappears because of an update, the automation would ensure that it is still possible to install the app, without requiring manual intervention from the maintainers.

flathub documents said that and this app uses extra data, so it is necessary for user can install even if I forgot or being late to merge the pr (the link will be cannot found soon after the upstream bump the version)

imngkhang:
> Wouldnt it make more sense, though, to ignore minor versions, so the flathub is slightly behind official releases, but is at least tested and not mystery package?
> 
> ive had stuff break with minor version bumps. I am not super comfortable with untested stuff going around. It is fine if not perfectly up to date

you don't know, they have mandatory updates, which disable all the functions of old versions, forces end users to update to the lastest version (which is bad), and ignoring minors is go againist the initial reason for me to made this wrapper: too tired to download it manually, and want an automatically update version of it. And the time between major and older major is 1-2 years, they can active the mandatory updates in this gap.

> flathub documents said that and this app uses extra data, so it is necessary for user can install even if I forgot or being late to merge the pr (the link will be cannot found soon after the upstream bump the version)

and I said that, their links to the old minor will be remove between hours, and it will cripple the ability to install the app for new users immediately

teamcons:
Make sure it works the couple first times pls

imngkhang:
ok I will check it

imngkhang:
I forgot to said that the deb version it follows the appimage structure, so breaks will be never occur, unless distros or the runtimes remove xwayland

PR #1118, opened 2026-08-17
Add finish-args-own-name-org.freedesktop.StatusNotifierItem-2-1 exception to md.obsidian.Obsidian

joshua-stone:
Related: https://github.com/flathub/md.obsidian.Obsidian/pull/556

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -5041,7 +5041,8 @@
         "stable": {
             "finish-args-has-socket-ssh-auth": "Predates the linter rule",
             "finish-args-home-filesystem-access": "Predates the linter rule",
-            "flathub-json-automerge-enabled": "Predates the linter rule"
+            "flathub-json-automerge-enabled": "Predates the linter rule",
+	    "finish-args-own-name-org.freedesktop.StatusNotifierItem-2-1": "Needed in order for the obsidian-tray plugin to work on KDE systems: https://github.com/flathub/md.obsidian.Obsidian/issues/555"
         }
     },
     "me.ahola.aphototoollibre": {


Erick555:
What's the reason no other app on flathub have this issue?

joshua-stone:
@Erick555 Frankly, I don't have an answer for why no other (Electron) app has this issue. [I even compared permissions in other Electron apps for enabling tray support](https://github.com/search?q=org%3Aflathub+tray&type=code), but I haven't found anything conclusive.

What I do know however is that this is documented behavior according to Electron:

https://github.com/electron/electron/blob/main/docs/api/tray.md

```
* Tray icon uses StatusNotifierItem by default, when it is not available in user's
  desktop environment the GtkStatusIcon will be used instead. If StatusNotifierItem is
  available, the first tray icon created will use SNI, while subsequently-created icons
  will use GtkStatusIcon.

* The click event is emitted when the tray icon receives activation from user, however
  the StatusNotifierItem spec does not specify which action would cause an activation,
  for some environments it is left mouse click, but for some it might be double left
  mouse click.

* In order for changes made to individual MenuItems to take effect, you have to call
  setContextMenu again. For example:
```

```javascript
const { app, Menu, Tray } = require('electron')

let appIcon = null
app.whenReady().then(() => {
  appIcon = new Tray('/path/to/my/icon')
  const contextMenu = Menu.buildFromTemplate([
    { label: 'Item1', type: 'radio' },
    { label: 'Item2', type: 'radio' }
  ])

  // Make a change to the context menu
  contextMenu.items[1].checked = false

  // Call this again for Linux because we modified the context menu
  appIcon.setContextMenu(contextMenu)
})

```

[The `setContextMenu` usage in the obsidian-tray plugin source code shares a lot in common with the upstream usage example:](https://github.com/dragonwocky/obsidian-tray/blob/ac353bb5a6e59fea9339befcf1c89fe0d4d97920/main.js#L188)

```javascript
  createTrayIcon = () => {
    destroyTray();
    if (!plugin.settings.createTrayIcon) return;
    log(LOG_TRAY_ICON);
    const obsidianIcon = nativeImage.createFromDataURL(
        plugin.settings.trayIconImage ?? OBSIDIAN_BASE64_ICON
      ),
      contextMenu = Menu.buildFromTemplate([
        {
          type: "normal",
          label: ACTION_QUICK_NOTE,
          accelerator: plugin.settings.quickNoteHotkey,
          click: addQuickNote,
        },
        {
          type: "normal",
          label: ACTION_SHOW,
          accelerator: plugin.settings.toggleWindowFocusHotkey,
          click: showWindows,
        },
        {
          type: "normal",
          label: ACTION_HIDE,
          accelerator: plugin.settings.toggleWindowFocusHotkey,
          click: hideWindows,
        },
        { type: "separator" },
        { label: ACTION_RELAUNCH, click: relaunchApp },
        { label: ACTION_CLOSE, click: closeVault },
      ]);
    tray = new Tray(obsidianIcon);
    tray.setContextMenu(contextMenu);
    tray.setToolTip(replaceVaultName(plugin.settings.trayIconTooltip));
    tray.on("click", () => {
      if (process.platform === "darwin") {
        // macos does not register separate left/right click actions
        // for menu items, icon should open menu w/out causing toggle
        tray.popUpContextMenu();
      } else toggleWindows(false);
    });
  };
```

Now, the docs mentioned "Tray icon uses StatusNotifierItem *by default*", which led me to investigating whether Obsidian would resort to a fallback mechanism depending on the detected desktop environment. I tried installing Fedora spins for KDE, LXQt, and COSMIC as those were all desktop environments with Wayland sessions as well as having support for tray icons, however the obsidian-tray icon didn't appear to use any detected fallback. That being said running Obsidian with `--talk-name=org.kde.StatusNotifierWatcher` and `--own-name=org.freedesktop.StatusNotifierItem-2-1` made the tray icon show up.

teamcons:
Ah, well,

joshua-stone:
Thanks for merging this. I really tried to find an alternative path due to the concerns around `--own-name=`, but this appears to be the only method that works for now.

teamcons:
yeah, i recognized your name, i know you know what you're doing

PR #1120, opened 2026-08-18
Add exception for Bottles UMU runtimes sharing with host

sambartik:
Follows the same path as Lutris and Heroic Launchers as both have this same exception here.

Complementing PR to enable it in Bottles flatpak manifest: https://github.com/flathub/com.usebottles.bottles/pull/548

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -2432,7 +2432,7 @@
     },
     "com.usebottles.bottles": {
         "*": {
-            "finish-args-contains-both-x11-and-wayland": "Predates the linter rule",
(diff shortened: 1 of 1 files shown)


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -2433,7 +2433,8 @@
     "com.usebottles.bottles": {
         "*": {
             "finish-args-contains-both-x11-and-wayland": "Predates the linter rule",
```
orowith2os:
```suggestion
            "finish-args-contains-both-x11-and-wayland": "App uses Wayland, and wine installations inside of it can use X11",
```

Forgot GitHub could do stuff like this. Would be good to change this at the same time, but not strictly necessary.

PR #1123, opened 2026-08-20
Add exception for Butler to restart itself

cassidyjames:
1.8.0 is adding a hardware acceleration setting, and for a good experience, the app prompts to restart itself. Afaiui this requires using `flatpak-spawn` to relaunch itself.

Removed the unused automerge exception as long as I was here.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -490,7 +490,7 @@
     },
     "com.cassidyjames.butler": {
         "stable": {
-            "flathub-json-automerge-enabled": "Predates the linter rule"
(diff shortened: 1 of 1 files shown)


teamcons:
flatpak-spawn pretty much is a "everything is now unsandboxed" permission. It is very extreme, only for the convenience of not restarting it manually.

CodedOre:
Have to agree. 

Wouldn't it be possible to restart the application inside the sandbox?

cassidyjames:
I think I misunderstood [Matthias's blog post](https://blogs.gnome.org/mclasen/2018/11/26/an-update-on-flatpak-updates/) about having apps restart themselves. 😅 I think I see what I need to do to get the app to restart reliable without adding this sandbox hole. Sorry for the noise!

teamcons:
may your beard remain shiny and sharp, Cassidy

PR #1124, opened 2026-08-22
Add COSMIC theme exception for io.github.marcelogomes90.FileCrate

marcelogomes90:
This adds a stable exception for `io.github.marcelogomes90.FileCrate`.

File Crate is a native libcosmic application. Its Flatpak requires read-only access to `~/.config/cosmic` to follow the host COSMIC theme. I tested the application without this permission, but it no longer matched the host theme.

Related Flathub submission: https://github.com/flathub/flathub/pull/9869

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -4059,6 +4059,11 @@
             "finish-args-own-name-wildcard-org.gnome.gimagereader": "Predates the linter rule"
         }
     },
+    "io.github.marcelogomes90.FileCrate": {
+        "stable": {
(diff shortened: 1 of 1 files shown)


teamcons:
Isnt it exposed through their toolkit?
If not it would be better to discuss it with them

marcelogomes90:
As far as I could determine, this is not currently exposed through the libcosmic toolkit for sandboxed applications. There are also existing applications with the same exception for read-only access to xdg-config/cosmic, which is why I followed this approach.

teamcons:
> As far as I could determine, this is not currently exposed through the libcosmic toolkit for sandboxed applications. 

Why dont they?
Is it planned?

marcelogomes90:
> > As far as I could determine, this is not currently exposed through the libcosmic toolkit for sandboxed applications. 
> 
> 
> 
> Why dont they?
> 
> Is it planned? 
> 
> 

I’m not part of the cosmic team, but I believe this is planned for the future. I’ll open an issue with them, and once it’s implemented, i’ll update the flatpak permissions to remove this access.

teamcons:
Tbh i was fine with granting the exception

PR #1127, opened 2026-08-23
Add host filesystem exception for io.github.marcelogomes90.FileCrate

marcelogomes90:
This adds a stable exception for io.github.marcelogomes90.FileCrate.

File Crate is an archive manager. Its requires host filesystem access because Wayland drag-and-drop delivers plain host paths that the document portal has never exported, and the file chooser portal cannot cover that case. I tested the application without this permission, but dropping an archive onto the window always failed, and the properties panel showed a /run/user/1000/doc/ path instead of the real archive location.

Reading and writing large archives through the document portal's FUSE mount is also noticeably slower than working on the real paths.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -4059,6 +4059,11 @@
             "finish-args-own-name-wildcard-org.gnome.gimagereader": "Predates the linter rule"
         }
     },
+    "io.github.marcelogomes90.FileCrate": {
+        "stable": {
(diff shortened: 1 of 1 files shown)


marcelogomes90:
Hi! Could you let me know why this PR was closed? Is there anything I need to change or address?

CodedOre:
The associated submission (https://github.com/flathub/flathub/pull/9869) was closed already. An exception PR without an submission or existing app has no real purpose.

marcelogomes90:
I had closed the PR myself so I could submit the correct exception request first. I was waiting for it to be accepted before opening a new pull request, so the new PR is already open. https://github.com/flathub/flathub/pull/10008

marcelogomes90:
Do I need to open a new PR here?

CodedOre:
Since the associated submission is closed for the moment (see https://github.com/flathub/flathub/pull/10008#issuecomment-5522089677), I close this for the time being.

PR #1131, opened 2026-08-25
exceptions: Add another exception to cx.modal.TestCenter

jonas2515:
Turns out "finish-args-host-var-access" is needed after all because we access /var/lib/extensions from within the sandbox.


jonas2515:
Closing, already done on master in the mean time, see https://github.com/flathub-infra/flatpak-builder-lint/commit/f47eac95d9796588d6d7d0c4f5ab2c31d863df1f

PR #1132, opened 2026-08-27
Add finish-args-login1-system-talk-name for org.waywallen.waywallen

hypengw:
Required to exit when user logout and stop renderer when user lockscreen


hypengw:
https://github.com/flathub/org.waywallen.waywallen/pull/19

teamcons:
Why not portals? Gtk integrates transparently a way to watch session changes, it would be surprising for KDE to not have something similar

PR #1135, opened 2026-09-01
exceptions: add finish-args-contains-both-x11-and-wayland for Git Cola

davvid:
Git Cola (Qt) needs `--socket=wayland`.
Gitk can be launched as a subprocess, and needs `--socket=x11`.

Related-to: https://github.com/flathub/com.github.git_cola.git-cola/issues/124

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -1045,6 +1045,7 @@
     "com.github.git_cola.git-cola": {
         "stable": {
             "appid-uses-code-hosting-domain": "app-id predates this linter rule",
+            "finish-args-contains-both-x11-and-wayland": "Needed for gitk subprocesses",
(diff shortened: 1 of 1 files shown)


teamcons:
whats gitk and why is there a second gui app inside the gui app

what does gitk devs think about needing wayland instead of x11

tbertels:
gitk is the Git repository browser https://git-scm.com/docs/gitk
It uses [tk](https://en.wikipedia.org/wiki/Tk_(software)), which doesn't work yet on Wayland: https://lwn.net/Articles/908636/ 

image

image

teamcons:
Please @tbertels pester the tk devs to consider a wayland backend

tbertels:
> Please @tbertels pester the tk devs to consider a wayland backend

It looks like they're working towards it already: https://github.com/tcltk/tk/commit/d466fbf80c66a52db0191f81467a0c6ccb8180ab

teamcons:
aw yea thank you

PR #1140, opened 2026-09-02
exceptions: add finish-args-contains-both-x11-and-wayland for org.dpsoftware.FireflyLuciferin

sblantipodi:
Requesting exception for `finish-args-contains-both-x11-and-wayland`

Firefly Luciferin genuinely requires both `--socket=x11` and `--socket=wayland` the app is unusable without either, confirmed via flatpak override testing:

Without X11 (--nosocket=x11, Wayland only): JavaFX/Glass fails to start entirely it has no native Wayland backend and always initializes through XWayland, even in a pure Wayland session:

```
Exception in thread "main" java.lang.UnsupportedOperationException: Unable to open DISPLAY
    at com.sun.glass.ui.gtk.GtkApplication.<init>(GtkApplication.java:153)
    at com.sun.glass.ui.gtk.GtkPlatformFactory.createApplication(GtkPlatformFactory.java:40)
    at com.sun.glass.ui.Application.run(Application.java:143)
    at com.sun.javafx.tk.quantum.QuantumToolkit.startup(QuantumToolkit.java:289)
    at com.sun.javafx.application.PlatformImpl.startup(PlatformImpl.java:281)
    ...
```

Without Wayland (--nosocket=wayland, X11 only): screen capture via PipeWire (pipewiresrc) fails to negotiate a supported buffer format and the pipeline errors out:

```
WARN pipewiresrc gstpipewiresrc.c:929:on_state_changed:<pipewiresrc0> error: stream error: no more input formats
WARN basesrc gstbasesrc.c:3187:gst_base_src_loop:<pipewiresrc0> error: Internal data stream error.
WARN basesrc gstbasesrc.c:3187:gst_base_src_loop:<pipewiresrc0> error: streaming stopped, reason not-negotiated (-4)
```

This is the app's core function (screen capture for ambient LED lighting), so it's not an optional feature we can gate behind X11-only support.

We tested dropping each socket individually (not just removing both) and confirmed each failure mode above with flatpak override --user. 

I commit to removing the x11 socket and keeping only the fallback-x11 socket as soon as JavaFX supports the Wayland backend.

Happy to provide further logs or test additional configurations if useful for review.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -5877,7 +5877,8 @@
     },
     "org.dpsoftware.FireflyLuciferin": {
         "stable": {
-            "finish-args-kwin-talk-name": "Predates the linter rule"
+            "finish-args-kwin-talk-name": "Used for StatusNotifierItem/AppIndicator and to detect NightLight status via DBUS.",
+            "finish-args-contains-both-x11-and-wayland": "JavaFX requires X11, GStreamer along with Pipewiresrc requires Wayland."
         }
     },
     "org.dune3d.dune3d": {


sblantipodi:
I add more context to the PR.
I'm the [author of the software](https://github.com/sblantipodi/firefly_luciferin).

This is the latest PR 
https://github.com/flathub/org.dpsoftware.FireflyLuciferin/pull/39.

CodedOre:
> I commit to removing the x11 socket and keeping only the fallback-x11 socket as soon as JavaFX supports the Wayland backend.

Is Wayland support for JavaFX still planned? AFAIK there is no momentum behind this.

Also, to get more context: \
What changed in the software that the exception is now needed, but wasn't before?

sblantipodi:
> Is Wayland support for JavaFX still planned? AFAIK there is no momentum behind this.

Yes, a pure Wayland backend is in progress:
https://bugs.openjdk.org/browse/JDK-8281970

I hope to see it in the near future, and I'll update this exception as soon as a full Wayland backend is available.

There is also some recent progress around this topic:
https://neatguycoding.com/posts/2026-05-18-javaone-2026-javafx-26-today/

> Also, to get more context: What changed in the software that the exception is now needed, but wasn't before?

The software is now using GPU accelerated GStreamer pipelines for screen capture.

Previously, I was using a CPU based pipeline:

`pipewiresrc fd={1} path={2} keepalive-time=PIPEWIRE_KEEPALIVE ! videorate drop-only=true ! videoscale ! videoconvert`

I have now switched to GPU accelerated pipelines, for example:

`pipewiresrc fd={1} path={2} keepalive-time=PIPEWIRE_KEEPALIVE ! vapostproc ! videorate drop-only=true`

and:

`pipewiresrc fd={1} path={2} keepalive-time=PIPEWIRE_KEEPALIVE ! videorate drop-only=true ! glupload ! glcolorscale ! glcolorconvert`

These GPU accelerated pipelines require Wayland support in the current application stack. Without the Wayland backend, the pipeline does not work at all.

Wayland is required independently of the UI backend because Luciferin uses the Wayland desktop session for screen capture through XDG Desktop Portal and PipeWire.

On a Wayland session, screen capture is exposed to applications through the XDG Desktop Portal ScreenCast interface. The portal obtains the screen content from the Wayland compositor and exposes it to the application as a PipeWire stream.

The change to GPU accelerated pipelines provides a significant performance improvement compared to the previous CPU based approach, which is why I would prefer to keep this implementation rather than revert to the CPU pipeline.

As you can see from the many releases over the years of development, my application is actively maintained and regularly updated. I will do my best to remove this exception as soon as a proper Wayland backend becomes available to JavaFX.

P.S. The latest release also brings full ARM64 support. This isn't related to this exception, but it is a highly requested feature (even from Flatpak owners some time ago) and one that I'm particularly happy to have finally delivered.

CodedOre:
Alright, I'm fine with this exception.

sblantipodi:
Thank you for your time CodedOre, appreciated!

PR #1141, opened 2026-09-03
exceptions: finish-args-flatpak-appdata-folder-org.onlyoffice.desktopeditors-data-onlyoffice-desktopeditors-recover-ro-access for page.wisha.platen

wishawa:
App submission PR: https://github.com/flathub/flathub/pull/10041.

When you copy an image object in OnlyOffice, the clipboard data contain a link to the location where OnlyOffice store the actual file. Platen needs to access these files, so that when users paste equations from OnlyOffice to Platen we can recover the formula.

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -8487,6 +8487,11 @@
             "finish-args-unnecessary-xdg-config-dconf-rw-access": "needed to access host dconf configuration at launch time"
         }
     },
+    "page.wisha.platen": {
+        "stable": {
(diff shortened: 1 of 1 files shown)


teamcons:
I am not against it, but why is there a feature specific to one exact random office suite?

wishawa:
It is a single feature (copy equation object in your slide, paste it in Platen to recover the formula) for three office suites (LibreOffice, OnlyOffice, and Google Slides). No extra permission is required for the other two because they store all the image data in the clipboard already (whereas OO stores a _url_ of the image file).

OnlyOffice is 100% not random. I'm pretty sure it is the second most popular office suite among Linux users. Look for example at the [AlternativeTo.net page on LibreOffice](https://alternativeto.net/software/libreoffice/), or the [comments on Reddit](https://www.reddit.com/r/pop_os/comments/1o7y8mm/alternative_to_libre_office/). I myself use OnlyOffice. It's great.

teamcons:
Accepting because it works for the other office suits. So the permission is more of a matter of avoiding the feature to break for users using OnlyOffice

PR #1142, opened 2026-09-04
Add exception for com.inochi2d.inochi-session

grillo-delmal:
This exception is needed so that the packaged version of the application is able to save the tracking configuration into the opened model file.


CodedOre:
As I already asked (see https://github.com/flathub/com.inochi2d.inochi-session/pull/7#issuecomment-5543883335):

Why is it not possible to use the FileChooser portal instead?

PR #1143, opened 2026-09-05
Add finish-args exception for Proton Pass

proletarius101:
Related: https://github.com/flathub/me.proton.Pass/issues/37

Similar to Bitwarden, Proton Pass uses the same arboard library, so it can perform the essential function of **copying**, which requires x11 in Flatpak, whereas the main Electron application requires Wayland

### DIFF
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -8800,5 +8800,5 @@
         "*": {
             "finish-args-contains-both-x11-and-wayland": "X11 access is needed for clipboard access"
         }
-    },
+    }
 }
--- a/flatpak_builder_lint/staticfiles/exceptions.json
+++ b/flatpak_builder_lint/staticfiles/exceptions.json
@@ -8798,7 +8798,7 @@
     },
     "me.proton.Pass": {
         "*": {
-            "finish-args-contains-both-x11-and-wayland": "X11 access is needed for clipboard access",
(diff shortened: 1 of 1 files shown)


proletarius101:
@CodedOre Could you please look into this?

CodedOre:
So, you'd essentially want X11 only to copy some content to the clipboard?

I'm sorry, but I can't see this being enough to justify adding this exception. 

You should find a proper solution that works on Wayland.

proletarius101:
@CodedOre In the same file, there is a precedence for the exact reason (same library, same dilemma): https://github.com/proletarius101/flatpak-builder-lint/blob/cbb1cbf3e34cd5d884e8a1f46c6fa54e523897a0/flatpak_builder_lint/staticfiles/exceptions.json#L414

Another maintainer @barthalion hinted that I should apply for the exception: https://matrix.to/#/!xFJJgxGKItHQPfstha:matrix.org/$5njH_vdIlUfQMn-fTekptgj8widPr0KRwzEK0qfG1Ds?via=kde.org&via=matrix.org&via=gnome.org

proletarius101:
> So, you'd essentially want X11 only to copy some content to the clipboard?

This is a password manager. If you cannot copy your username/password, it's basically useless.

barthalion:
I have agreed to these exceptions as the upstream issue has been reported at https://github.com/1Password/arboard/issues/223. Whether it will ever be looked at is a different question.

PR #1144, opened 2026-09-05
Add exceptions for io.github.michelet76.CosmicExtCheatsheet

michelet76:
Exception request for COSMIC Cheatsheet, submitted at flathub/flathub#10088. I am the author.

The app shows every keyboard shortcut the COSMIC desktop currently has configured, read live from the compositor's own configuration.

- **finish-args-host-os-ro-filesystem-access** — COSMIC ships its keyboard shortcut defaults as a data file in `/usr/share/cosmic` on the host. Reading them is the entire purpose of the app, and Flatpak reserves `/usr` inside the sandbox, so read-only host-os access is the only route; `--filesystem=/usr/share/cosmic:ro` is refused with "Path /usr is reserved by Flatpak". `XDG_DATA_DIRS` is extended with `/run/host/usr/share` so the configuration library finds the file at its host location. Without it the app can only show the user's custom bindings. Nothing on the host is written.
- **finish-args-unnecessary-xdg-config-cosmic-rw-access** — the global shortcut is registered by writing a single entry into the compositor's custom shortcut file under `~/.config/cosmic`, which cosmic-comp watches and reloads immediately. This is the same mechanism COSMIC Settings uses, and xdg-desktop-portal-cosmic has no GlobalShortcuts interface, so there is no portal alternative. The app only ever adds, updates or removes its own entry, re-reads the file immediately before writing, and refuses to write at all if any entry fails to parse, so other shortcuts cannot be lost.
- **finish-args-flatpak-spawn-access** — a row in the cheatsheet can be clicked to run the shortcut it documents, for example opening a terminal or changing the volume. Those commands come from the compositor's own system-action configuration and belong to the host session; they do not exist inside the sandbox, so they are handed to `flatpak-spawn --host`. Only commands already present in the user's COSMIC shortcut configuration are ever executed.


teamcons:
grafik

PR #1147, opened 2026-09-06
Add desktop-file-is-nodisplay entry to fcitx

wengxt:
Setting NoDisplay It's a request from KDE upstream: https://invent.kde.org/kde-linux/kde-linux/-/work_items/743

It seems to make sense to me because input method is a thing that user doesn't start manually from menu, even though there's GUI. However desktop file is still required to set certain metadata field to allow KDE to display it in another place and grant required permission.

I'm aware that doc says this is never granted but maybe this is still a meaningful case: https://docs.flathub.org/docs/for-app-authors/linter#desktop-file-is-nodisplay

Outside the flatpak world,

E.g. https://github.com/ibus/ibus/blob/main/ui/gtk3/ibus-ui-wayland.desktop.in , ibus also Set NoDisplay=true in their desktop file


CodedOre:
To be honest, this seems like the kind of software that may not be elible for inclusion with the current policies. 

But I do see the argument for it and its on Flathub already. So I merge it.

PR #1148, opened 2026-09-06
exceptions: add home filesystem access for io.github.killown.flux

killown:


teamcons:
Granted, as it was offered as compromise against host, for a file manager. As those are uncommon we can accept it

PR #1149, opened 2026-09-07
Add exceptions for uk._92li.lingyicute.ComputeSec

lingyicute:
# Request for Exception

**App ID:** `uk._92li.lingyicute.ComputeSec`
**Requested exception repo:** `flathub` (stable)

---

## Summary

ComputeSec is a **security hardening assessment tool**. Its core mission is to help users turn a Linux machine into a trustworthy computing environment by evaluating firmware (HSI), kernel hardening, hardware vendor reputation, and usage habits.

**X11 is fundamentally incompatible with this mission.** We explicitly do **not** support X11 and request an exception to ship Wayland-only (`--socket=wayland` without `fallback-x11`/`x11`).

---

### 1. X11 is insecure by design

- No isolation: any X client can **keylog, screenshot, inject input, read clipboard, enumerate windows** of any other client
- No sandbox: `X11` permission effectively breaks Flatpak sandbox for input and display
- For a tool that warns about `2.4G wireless receivers (MouseJack)`, `NTFS/Windows contamination`, `unencrypted swap`, and `IOMMU disabled`, running on X11 would be **self-contradictory** – the display server itself would be the biggest vulnerability, and our score would be meaningless

### 2. GNOME has dropped X11

- GNOME 49+ is **Wayland-only** (https://www.phoronix.com/news/GNOME-49-Wayland-Only)
- Fedora Workstation 41+, RHEL 10, Ubuntu 25.10+ default to Wayland-only, no Xorg session installed
- `org.gnome.Platform` 50 runtime is Wayland-first; libadwaita apps are designed for Wayland
- ComputeSec uses GTK4 / libadwaita 1.4+ with no X11-specific code paths

### 3. Security software must lead, not lag

- As a security assessment tool, we must **refuse to run in an insecure environment** rather than silently degrade. This is similar to other security-focused apps that are Wayland-only

---

## Why `system-folder-ro-access` and `user-folder-ro-access` are safe (for context)

The other two linter errors you see are:

- `finish-args-flatpak-system-folder-ro-access` – `/var/lib/flatpak:ro`
- `finish-args-flatpak-user-folder-ro-access` – `~/.local/share/flatpak:ro`

These are **read-only** and used only to count installed Flatpak apps (excluding runtimes `org.gnome.*`, `org.kde.*`, `org.freedesktop.*`, etc.) to evaluate sandbox adoption – a core feature on the "Usage Habits" page.

- `:ro` – no write
- Gracefully degrades to "unknown" if revoked via Flatseal
- Less invasive than the alternative `--talk-name=org.freedesktop.Flatpak` which allows `flatpak-spawn --host` arbitrary command execution (= sandbox escape) and was **removed** in v1.0.3 for privacy reasons
- Similar permissions are granted to Warehouse, Flatseal, and other Flatpak management tools

We request exceptions for these as well, with justification:

> Count installed Flatpak apps (excluding runtimes) to evaluate sandbox adoption – read-only, gracefully degrades

---

## Conclusion

- X11 is loose: any app can snoop on any other app – this is the opposite of what a security tool should allow
- `system-folder:ro` is strict in linter but is actually safe (read-only, counted, not executed)
- We believe security software should **refuse** to run on insecure foundations, not silently support them

If Flathub cannot grant Wayland-only exception, we would rather not be listed than ship with X11 support that contradicts our security mission.

Thank you for consideration.

— lingyicute, author of ComputeSec
https://github.com/lingyicute/ComputeSec
https://s.92li.uk/


teamcons:
AI;DR

teamcons:
Also... The fallback is xwayland... It doesnt get privileged by the compositor...

lingyicute:
？？？

It is profoundly ironic that you question my competence while fundamentally misunderstanding Flatpak's own permissions.

Per your own documentation (https://docs.flatpak.org/en/latest/sandbox-permissions.html):
fallback-x11: "Show windows using X11, if Wayland is not available."

And since XWayland exists, why can't display be handled via Wayland? This just means if a user is running a bare X.org host session, my application would fallback to connecting directly to the native X11 socket. As a security developer, I strictly refuse to allow my auditing tool to run in a native X11 session, which fundamentally lacks GUI isolation and allows global keylogging. Your assumption that this permission only implies Xwayland is technically incorrect.

Between your failure to understand your own documentation and your toxic "AI;DR" gatekeeping, it's clear Flathub is not a suitable platform for security-focused tools.

Okay. Keep the PR closed. Goodbye.

teamcons:
Not what im saying but ok

PR #1150, opened 2026-09-07
exceptions: Add exception for org.pwsafe.pwsafe

froggie2523:
Related to: https://github.com/flathub/org.pwsafe.pwsafe/issues/7 or https://github.com/flathub/org.pwsafe.pwsafe/issues/26.

Currently, the org.pwsafe.pwsafe app uses both "--socket=wayland" and "--socket=fallback-x11" in the finishing section.
For the Autotype feature, it needs "--socket=x11" instead of "--socket=fallback-x11".


CodedOre:
Not a great fan of this to be honest.

Not only is the request to activate an legacy socket with less security, it's also for the intended purpose of letting the application type into other applications.

I would argue this is better left for the users to override manually if they understand it. Utlimately, getting a portal for that purpose would be better to pursue.

froggie2523:
Thanks for your feedback. The app has been updated to provide end users with information on how to manually override the restriction if needed. This PR can be closed.

PR #1156, opened 2026-09-10
Add finish-args-only-wayland exception for io.github.aradar46.Dictator

aradar46:
Dictator is a push-to-talk dictation app for GNOME. It does not support X11, so its manifest ships `--socket=wayland` with no X11 fallback.

The reason is not a preference. The app binds a system-wide shortcut through `org.freedesktop.portal.GlobalShortcuts`, and that call needs a Wayland surface handle exported with `export_handle()` to associate the shortcut with the window. There is no X11 code path in the app.

It refuses to run on anything else rather than degrading ([source at v0.1.0](https://github.com/aradar46/dictator/blob/v0.1.0/src/dictator/dictator.py)):

```python
surface = self.window.get_surface()
if isinstance(surface, GdkWayland.WaylandToplevel):
    surface.export_handle(lambda s, h, d: self.portal.bind("wayland:" + h), None)
else:
    self.dictation.fail("Dictator needs a GNOME Wayland session.")
```

Adding `--socket=fallback-x11` would request a socket the app can never use, so I would rather not ask for it.

Repository: https://github.com/aradar46/dictator


aradar46:
Withdrawing this for now, the app isn't ready for submission yet

PR #1161, opened 2026-09-11
Update exceptions for org.telegram.desktop and its forks

ilya-fedin:
https://github.com/flathub/org.telegram.desktop/pull/1111


teamcons:
Can you explain to me why you need it please?

ilya-fedin:
> Can you explain to me why you need it please?

Because updating commits by hand in two places in upstream repo and then third time in flathub would be a nightmare

teamcons:
Okay sounds reasonable

teamcons:
enjoy your freedom

PR #1164, opened 2026-09-13
Add automerge exception for Super Productivity

johannesjo:
Request `flathub-json-automerge-enabled` for `com.super_productivity.SuperProductivity`.

Super Productivity is a verified app maintained by its upstream author. Its only `x-checker-data` source tracks the stable `superProductivity-amd64.deb` from upstream GitHub releases. The latest 20 stable releases are immutable, and GitHub exposes SHA-256 digests for their assets. Flathub CI will continue to gate every update PR.

The recent cadence is 26 checker PRs since 2026-04-01 (about 1.1 per week, including two superseded PRs). We understand that a successful build does not prove the app launches and will continue monitoring releases and revert or disable automerge if an update breaks the Flatpak.

Policy: https://docs.flathub.org/docs/for-app-authors/maintenance#automatically-merging-updates

Manifest: https://github.com/flathub/com.super_productivity.SuperProductivity

Verification: https://flathub.org/apps/com.super_productivity.SuperProductivity


CodedOre:
The exception is usually only given for few applications, mostly when `extra-data` is involved. And to be frank, I don't see why an productivity checker needs to get an update weekly at first and secondly as fast as possible that the manual check can be ignored.

PR #1165, opened 2026-09-13
Add exception for io.github.TeamWheelWizard.WheelWizard

matellush:
Because upstream has added support for Wayland for Wheel Wizard itself (https://github.com/TeamWheelWizard/WheelWizard/commit/9b130e1b22118c1bc40102e15f7c2c828ba23f30), we would like to add the `--socket=wayland` `finish-arg`. Sadly, the reason we cannot directly use `--socket=fallback-x11` with it is that the Dolphin Emulator currently does not properly support using Wayland and forces the use of X11. We hope that this will change in the future or that we could maybe patch in its Wayland support so we could use `fallback-x11` in the future, but right now we cannot easily support Wayland-only setups for all features of the application.

I've also updated the descriptions for the previous `finish-args-flatpak-appdata-folder-access` exception (now split into `finish-args-flatpak-appdata-folder-org.DolphinEmu.dolphin-emu-config-dolphin-emu-create-access` and `finish-args-flatpak-appdata-folder-org.DolphinEmu.dolphin-emu-data-dolphin-emu-create-access` due to the migration) in commit 4ad394743f439069c873566971943bdb0c8f7245.

Lastly, due to not requiring `finish-args-flatpak-spawn-access` anymore, I removed it with commit b688fc81a150eaebdc7c0a920e9950b0ab87badb. I saw that the automated PR #1162 already includes the removal of that stale exception, but since I figured that this PR would likely cause a merge conflict anyway, I removed the exception here.


flatpak_builder_lint/staticfiles/exceptions.json
```diff
@@ -3451,9 +3451,9 @@
...
-            "finish-args-flatpak-appdata-folder-org.DolphinEmu.dolphin-emu-data-dolphin-emu-create-access": "this program needs to be able to access the Dolphin Flatpak's config and data directories for configuration and Mii/mod management",
-            "finish-args-flatpak-spawn-access": "this program needs to be able to run host commands as specified in the manifest's wrapper scripts to support Flatpak and native Dolphin installations",
+            "finish-args-contains-both-x11-and-wayland": "this program supports Wayland but also depends on the Dolphin Emulator, which currently forces the use of X11, so not having X11 would break regular usage of the application",
+            "finish-args-flatpak-appdata-folder-org.DolphinEmu.dolphin-emu-config-dolphin-emu-create-access": "this program needs to be able to access the Dolphin Flatpak's config directory for configuration management",
+            "finish-args-flatpak-appdata-folder-org.DolphinEmu.dolphin-emu-data-dolphin-emu-create-access": "this program needs to be able to access the Dolphin Flatpak's data directory for Mii/mod management",
```
teamcons, line 3444: What if the user doesnt have it installed?
Also flathub apps shouldnt really communicate with each other. Theyre supposed to exist in isolation.

matellush: Which is why we bundle the Dolphin Emulator now. I am working on making Dolphin a Flatpak extension, but right now our manifest always has bundled Dolphin installed as a dependency. The issue is that Wheel Wizard defaults to the bundled Dolphin Emulator, as this emulator has been the main way to play the game. If the emulator itself does not work with X11, then enforcing `fallback-x11` would currently break the point of this application -- launching the Mario Kart Wii mod.

If you were referring to the Dolphin config/appdata access: Not granting these filesystem permissions would be detrimental to the user experience, since it appears that portal-based Dolphin "user folder" access can significantly impact/harm performance. Additionally, our users need it to automatically work with the native Dolphin or Flatpak Dolphin installation because they have their Wii NAND with saves, Miis, etc. stored there.

matellush: Due to having `:create`, the default setup flow does not even require having the Dolphin Flatpak installed in addition to Wheel Wizard. If users were to then use the Dolphin Flatpak, all saves would be already there.

matellush: I would like to emphasize that the point of this PR is only the Wayland/X11 issue, the redundant `flatpak-spawn` was dropped because it is not needed anymore (and should be automatically dropped by your stale exceptions checker anyway) and I just renamed the text in the other (already previously granted) exceptions because you migrated them and the description was not entirely accurate anymore.

teamcons: > Additionally, our users need it to automatically work with the native Dolphin or Flatpak Dolphin installation because they have their Wii NAND with saves, Miis, etc. stored there.

What if the user doesnt have it installed?

matellush: > > Additionally, our users need it to automatically work with the native Dolphin or Flatpak Dolphin installation because they have their Wii NAND with saves, Miis, etc. stored there.
> 
> What if the user doesnt have it installed?

As I said, it will default to using `~/.var/app/org.DolphinEmu.dolphin-emu/{config,data}/dolphin-emu` since our manifest has `:create` on these directories and these directories will exist when first launching Wheel Wizard. The user is free to choose a different directory, though.

matellush: It works just fine even if Dolphin is not installed separately.

PR #1168, opened 2026-09-14
Add app.liten.Gram exceptions

helinoscool:
- finish-args-home-filesystem-access: "To launch common paths from the terminal
 - finish-args-flatpak-spawn-access:  To access debuggers, linters and other programming tools


teamcons:
those are too broad permissions. The latter one is a full sandbox escape.
I get it is for an IDE and needs a lot to work, but that is covered under: https://docs.flathub.org/docs/for-app-authors/requirements#submissions-incompatible-or-restricted-by-flatpak-sandbox

Sorry :/

helinoscool:
The [owner](https://codeberg.org/GramEditor/gram/issues/449#issuecomment-22773910) wants me to support the IDE as a Flatpak. But if you do not accept it as the policy allows I understand. Is there also a way to make the IDE more compatible with the sandbox which would suffice?

BTW the home permission is not needed for people using the GUI. For which the document portal works well.

teamcons:
to not rely on both
and while we can grant the first one if we have to (which is not the case here),
we cannot grant finish-args-flatpak-spawn-access

helinoscool:
So if Gram works without flatpak-spawn-access it would be allowed? And I assume that using filesystem-host is as bad as the current situation? If so I will try looking for alternatives. Thank you for the help

teamcons:
> So if Gram works without flatpak-spawn-access it would be allowed? 

Yes. But that quite aggressively neuters the IDE since everything will be contained within the sandbox, resources and all. Which is why it is listed as a poor match for flatpak.


> And I assume that using filesystem-host is as bad as the current situation?

Nope. We really prefer portals/filechoosers. We can grant for an xdg folder like Projects or Documents, but would rather avoid if possible. Host and Home are nuclear options. Flatpak Spawn is Thanos level of no.

PR #1169, opened 2026-09-15
Add exception for xyz.froststrap.Froststrap

blazebsc:
* Add finish-args exception for Froststrap
Related: https://github.com/flathub/flathub/pull/10229
Froststrap installs, updates and launches the Sober Flatpak on the host via flatpak-spawn.


CodedOre:
No, that will not be granted.

PR #1171, opened 2026-09-15
Add exception for io.github.foxstudio201.dinoisekai

foxstudio-201:
Add exception for io.github.foxstudio201.dinoisekai


CodedOre:
There is neither an application or submission for that app id, nor is there a valid reason for that exception.

PR #1172, opened 2026-09-17
Add exceptions for org.chipass.ChiPass

whitequark:
See https://github.com/flathub/flathub/pull/10223


teamcons:
granted yo

whitequark:
Thanks!

PR #1174, opened 2026-09-21
Add finish-args-unnecessary-xdg-config-gtk-3.0-ro-access for io.gitlab.librewolf-community

any1here:
Required to get theme detection working.

https://bugzilla.mozilla.org/show_bug.cgi?id=1990196


CodedOre:
Ugh, this seems to me a lot like an workaround since Firefox is not using the proper way of reading accent colors. 

Using a GTK theme to read accent color is really not ideal. The Settings portal was added for a reason. 

Its not that big of a deal, so I'll grant it, given Firefox also already uses it. Just note that this needs a proper fix.

PR #1175, opened 2026-09-21
Add exception rule request for `com.p1security.diagng`

p1-mmr:
Hello,

This is a exception request rule linked to the following Flathub package add request: https://github.com/flathub/flathub/pull/10329

As said in the aforementioned pull request,

> DiagNG is a general purpose client for the diagnostic interface of Qualcomm Snapdragon basebands, present on a large part of Android phones and various USB modems.

And

> This application being a security research and pedagogy tool performing low-level interaction with peripheral hardware, it has then [design principles](https://github.com/P1sec/DiagNG#technical-architecture) where some of the required functionality (connecting to raw USB and privileged TTY pseudo-serial ports, also detecting and watching for these devices) is privileged and has been moved, in order to integrate gracefully to the Linux environment, to a [Polkit-authorized system DBus daemon, written in Rust](https://github.com/P1sec/DiagNG/tree/main/diagmond).<br><br>For this reason, a joint exception request for the `finish-args-flatpak-spawn-access` Flatpak rule will be submitted to the [`flathub-infra/flatpak-builder-lint`](https://github.com/flathub-infra/flatpak-builder-lint) repository along with this pull request so that this software, which was designed to be Flatpak-first, can use the [`flatpak-spawn`](https://github.com/search?q=repo:P1sec/DiagNG+%22flatpak-spawn%22&type=code) command.

In DiagNG, the [`flatpak-spawn`](https://github.com/search?q=repo:P1sec/DiagNG+%22flatpak-spawn%22&type=code) command is used to launch the [`diagmond`](](https://github.com/P1sec/DiagNG/tree/main/diagmond)) daemon, which is a system D-Bus service exposing the privileged actions related to the app. For each individual privileged action, a Polkit authorization dialog is displayed to the user, so the usual Linux desktop right models is harmoniously respected.

It is also used to check the presence of the Google Android Tools [ADB](https://en.wikipedia.org/wiki/Android_Debug_Bridge) daemon on the system, and the version of Wireshark on the main system.

Like in [QCSuper](https://github.com/P1Sec/QCSuper), piping a PCAP/GSMTAP protocolar stream to Wireshark is a key feature of DiagNG, hence we also need `flatpak-spawn` to spawn an instance of Wireshark which then receives a stream of PCAP data through an UNIX pipe, as displayed in the [demo video](https://www.youtube.com/watch?v=aMbaXFxMELw).

Regards,


orowith2os (commented):
LGTM

teamcons:
This just completely breaks the sandbox open.
I can think of several items under impermissible submissions this would fall into: https://docs.flathub.org/docs/for-app-authors/requirements#impermissible-submissions

flathub apps shouldnt rely on anything from the host being present. They shouldnt need anything privileged. They shouldnt need to just disregard the entire point of the flatpak format to function.

p1-mmr:
Well on my side I stumbled upon the OpenTrace project (https://github.com/flathub/io.github.Archeb.opentrace) which seemed to use `flatpak-spawn` to do something similar, aka running networking functionality with elevated privileges.

Regarding why using the mere USB portal is not possible for the needs of this app, I have provided explanations here in the course of the review process: https://github.com/flathub/flathub/pull/10329#discussion_r4065045730

There is no strict reliance on Wireshark or the ADB daemon not being present on the host, the program gracefully degrades when it is not the case: https://github.com/flathub/flathub/pull/10329#discussion_r4072178547

When Wireshark is present, it is interfaced with only in standard ways (sending a stream of PCAP - a format standardized enough to have a RFC - to stdin, and optionally installing a Lua plug-in - the Wireshark Lua API is pretty stable and constant and has been for a long time).

When the ADB daemon is present, it communicates with it through the ADB local protocol which is also standardized (https://cs.android.com/android/platform/superproject/main/+/main:packages/modules/adb/docs/dev/).

Regards,

p1-mmr:
> disregard the entire point of the flatpak

I sincerely have done my best to keep whatever can be kept in the sandbox: the vast majority of the program runs sandboxed, benefits of a recent GTK/Adwaita platform in a portable environment (one of the points of Flatpak), uses portals when portals can be used.

The daemon containing the HW interaction layer uses D-Bus + Polkit with individual authorization on each privileged USB/SPI access, so that it is contained at best and launching the privileged daemon through Polkit also goes through an explicit and informed [authorization request](https://github.com/P1sec/DiagNG#technical-architecture) to the user.

Querying the UDev devices present on the OS is something that stays in the Python code/sandbox. Querying ModemManager status through D-Bus stays in the sandbox too. Communicating with the ADB daemon socket stays in the sandbox scope too (it's a TCP-based protocol).

Regards,

teamcons:
> > disregard the entire point of the flatpak
> 
> I sincerely have done my best to keep whatever can be kept in the sandbox: the vast majority of the program runs sandboxed, benefits of a recent GTK/Adwaita platform in a portable environment (one of the points of Flatpak), uses portals when portals can be used.

Yes and i commend your efforts, and i am really sorry to be the grumpy guardian here
this one is like every permission in the universe rolled in together but worse. It breaks the sandbox.

i get flathub makes it super easy to distribute everywhere and be visible, but if we accept you we accept every system utility that breaks the sandboxing but want to be on the sandboxed software store. And then the point of flathub isnt even flatpak anymore

opentrace would be rejected on the same ground if it were submitted today and not in 2017

You can still distribute a flatpak bundle for people to install easily, or rpm and deb, but this rely on too low level stuff

@CodedOre what would you think here?

CodedOre:
I do have to agree with teamcons here. 

This application requires a bit too much system integration that makes it not the best fit together with Flatpak ATM. 

Especially if major functionality is degraded if host components are missing. 

I can see the benefit of an Flatpak in it's own repository, but I don't see it a good fit for Flathub.

PR #1176, opened 2026-09-22
Add OpenRCT2 read-only access to Flatpak Steam installations of RollerCoaster Tycoon

rmnvgr:
Hello,

OpenRCT2 added detection of Flatpak Steam installations of RollerCoaster Tycoon Classic/Deluxe/2: https://github.com/OpenRCT2/OpenRCT2/pull/27097

This exception would allow this detection to work in Flatpak OpenRCT2.


CodedOre:
These exceptions would only work for a native Steam install and to Role Coaster Tycoon 2 being installed into the default Steam installation folder. 

What happens, for example, if I'd had installed RCT on an external drive? What if I use the Flatpak version of Steam?

I understand the motivation behind this, but I'm not too eager to add a ton of permissions and the added exceptions for it.

rmnvgr:
> These exceptions would only work for a native Steam install and to Role Coaster Tycoon 2 being installed into the default Steam installation folder.

I believe these exceptions are for the Flatpak Steam directories? Flatpak OpenRCT2 already has static permissions for the native Steam.

> What happens, for example, if I'd had installed RCT on an external drive?

OpenRCT2 only checks the default paths, if it doesn't find a RCT installation, it will ask the user to manually point to it.

CodedOre:
Seems like OpenRCT2 does already have the native Steam paths. Thought those paths would need an exception as well.

Very well, I'll grant the exception.

rmnvgr:
Thank you!

